PR #4, opened 2020-09-30
Fix/docker compose network

rhyslbw:
- add ENV to select `cardano-node` config
- default to _mainnet_
- use _projects_ to naturally handle data volume grouping
- make Ogmios port configurable
- enable logging for `cardano-node` container
- add build info to README, and fix codestyle violations

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,14 @@
 **Ogmios** is a translation service written in Haskell running on top of [cardano-node](https://github.com/input-output-hk/cardano-node/). It offers a [JSON-WSP](https://en.wikipedia.org/wiki/JSON-WSP) interface through WebSockets and enables clients to speak [Ouroboros' mini-protocols](https://hydra.iohk.io/build/1070091/download/1/network.pdf#chapter.3) via remote procedure calls over JSON.
 
 ```
-            CARDANO RELAY                                     OGMIOS SERVER   
-                                                                           
-                  ◦                                                _=_       
-<-------->    ◦   ○   ◦                                          q(-_-)p     
-               ○ ◯ ◯ ○                                            _\=/_      
+            CARDANO RELAY                                     OGMIOS SERVER
+
+                  ◦                                                _=_
+<-------->    ◦   ○   ◦                                          q(-_-)p
+               ○ ◯ ◯ ○                                            _\=/_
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -5,25 +5,27 @@ services:
...
     restart: on-failure
     environment:
-      - OGMIOS_NETWORK=${NETWORK}
+      - OGMIOS_NETWORK=${NETWORK:-mainnet}
     command: ["--host", "0.0.0.0", "--node-socket", "/ipc/node.socket"]
     volumes:
       - node-ipc:/ipc
     ports:
-      - 1337:1337
+      - ${OGMIOS_WS_PORT:-1337}:1337
```
KtorZ, line 27: Note that this port is used for both the websocket and also "classic" http requests (like `/health`, `/benchmark.html` or `/ogmios.wsp.json`). I'd rather call that one simply `OGMIOS_PORT` :+1:

docker-compose.yml
```diff
@@ -5,25 +5,27 @@ services:
...
 
 volumes:
-  node-mainnet-db:
-  node-testnet-db:
```
KtorZ, line 28: I used separate volumes here originally to avoid overwriting existing volumes when switching from mainnet to testnet and vice-versa. Ideally, I'd like to write `node-{$NETWORK}-db:` but that's not (yet?) supported.

docker-compose.yml
rhyslbw: Will generalise

docker-compose.yml
rhyslbw: The suggestion here is to provide a project name to set the scope at the stack level, enabling simultaneous operation of networks, and a clean boundary when it comes time to clean things up.

``` console
NETWORK=testnet docker-compose -p cardano-ogmios-testnet up
```
This way it's possible to manage the stack independently, eg

``` console
docker-compose -p  cardano-ogmios-testnet down -v
```

README.md
```diff
@@ -87,21 +87,28 @@ to display, in-real-time, the current tip of MainNet.
 :point_up: This will run and connect together:
 
 - An OBFT Byron cardano-node, connected to `MainNet`.
-- An Ogmios server, listening to `localhost` on port `:1337`.
+- An Ogmios server using the `latest` Dockerhub build, listening to `localhost` on port `:1337`.
+- To build Ogmios from source, pass the `--build` flag
```
KtorZ, line 91: :+1:

KtorZ (approved):

PR #29, opened 2021-04-22
Introduce TypeScript client

rhyslbw:
Introduces an almost complete client implementation and two additional packages as detailed in the READMEs. This PR contains a bulk of code, however the changes are well isolated in commits with what should be enough detail in the messages to review each commit in sequence. The tests will be improved when a dependent project makes it easy to setup a local test network with concrete assertions, with the full e2e scope tested. The CI workflow has been extended to build, lint, and package the REPL, however tests are not yet running due to the dependency on the public testnet and lack of desire to manage a node on it.

_Not included, but plan to add:_
- Handle `Fault` response
- Health check method


clients/TypeScript/packages/client/src/TxSubmission/submitTx.ts
```diff
@@ -0,0 +1,127 @@
...
+              const { eraMismatch } = SubmitFail
+              return reject(new EraMismatchError(eraMismatch.queryEra, eraMismatch.ledgerEra))
+            } else if (errors.byron.UtxoValidation.assert(SubmitFail)) {
+              return reject(new errors.byron.UtxoValidation.Error(SubmitFail))
+            } else if (errors.byron.TxValidation.assert(SubmitFail)) {
+              return reject(new errors.byron.TxValidation.Error(SubmitFail))
+            } else if (Array.isArray(SubmitFail)) {
+              reject(SubmitFail.map(failure => {
+                if (errors.shelley.InvalidWitnesses.assert(failure)) {
+                  return new errors.shelley.InvalidWitnesses.Error(failure)
```
rhyslbw, line 34: I couldn't find a more elegant way of handling these [user-defined type guards](https://levelup.gitconnected.com/user-defined-type-guards-in-typescript-fad639e4944f)

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -0,0 +1,97 @@
...
+    }) => void
+    rollForward: (response: {
+      block: Block,
+      tip: Tip,
+      reflection: Mirror
+    }) => void
+  }) => void
+  requestNext: (options?: { mirror?: Mirror }) => void
+  shutdown: () => Promise<void>
+}
```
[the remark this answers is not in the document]
KtorZ: ( https://github.com/KtorZ/cardano-ogmios/issues/30 )

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -0,0 +1,97 @@
...
+export const createChainSyncClient = async (options?: {
+  connection?: ConnectionConfig
+}): Promise<ChainSyncClient> => {
+  return new Promise((resolve, reject) => {
+    createClientContext(options).then(context => {
+      const { socket } = context
+      socket.once('error', reject)
+      socket.once('open', async () => {
+        const initialIntersection = await findIntersect(
+          [await createPointFromCurrentTip({ socket, closeOnCompletion: false })],
```
KtorZ, line 42: I wish there was a special `tip` point. I've been considering implementing it just in Ogmios, but having this in the client is perhaps preferable.

clients/TypeScript/packages/client/src/ChainSync/findIntersect.ts
```diff
@@ -0,0 +1,40 @@
...
+// type Intersection = Ogmios['FindIntersectResponse']['result']['IntersectionFound']
+export type Intersection = { point: Point, tip: Tip }
+
+export const findIntersect = (points: Point[], context?: InteractionContext): Promise<Intersection> => {
+  return ensureSocket<Intersection>((socket) => {
+    return new Promise((resolve, reject) => {
+      const requestId = nanoid(5)
+      socket.once('message', (message) => {
+        const response: Ogmios['FindIntersectResponse'] = JSON.parse(message)
+        if (response.reflection.requestId !== requestId) { return }
```
KtorZ, line 16: I am not sure to get the rationale of not exposing the mirror/reflection to callers for `findIntersect`, but using it internally only :thinking: 

This is also discarding messages that are not intersection responses which may accidentally lead to messages being skipped? One thing to know about `FindIntersect` is that it is not allowed to interleave it within some `RequestNext` messages; this will lead to a client error from Ogmios and isn't possible to express using the Ouroboros mini-protocols (but possible with WebSockets unfortunately).

clients/TypeScript/packages/client/src/Connection.ts
```diff
@@ -0,0 +1,59 @@
...
+  socket?: WebSocket
+  connection?: ConnectionConfig
+  closeOnCompletion?: boolean
+  connectionString?: string
+}
+
+export type Mirror = { [k: string]: unknown }
+
+const createConnectionString = (connection?: ConnectionConfig): InteractionContext['connectionString'] =>
+  `ws://${connection?.host || 'localhost'}:${connection?.port || 1337}`
```
KtorZ, line 18: It should eventually be allowed to connect to a WebSocket through TLS; thus making the protocol part of the `ConnectionConfig` with a sensible default perhaps :thinking: ?

clients/TypeScript/packages/client/src/TxSubmission/errors.ts
```diff
@@ -0,0 +1,478 @@
...
+        (item as ProtocolVersionCannotFollow).protocolVersionCannotFollow !== undefined,
+      Error: class ProtocolVersionCannotFollowError extends CustomError {
+        public constructor (rawError: ProtocolVersionCannotFollow) {
+          super()
+          this.message = JSON.stringify(rawError.protocolVersionCannotFollow, null, 2)
+        }
+      }
+    }
+  }
+}
```
KtorZ, line 478: Doh :facepalm: ... I was wondering why this didn't cover Allegra and Mary errors... and realized I completely forgot to add them in the JSON-schema (and in the corresponding test code, which actually only ever generate Shelley errors, so this was never caught ...). 

There's more... because Allegra and Mary are a bit bigger than Shelley, and while most constructors are the same, enough are different enough to make it quite tricky to handle in a generic manner... I'll add them A.S.A.P to the schema & test generator.

KtorZ: Okay, fixed in: 773219e44689a9dbf4a72b4a58571c198e52fd69 

I've added a summary of the changes between Shelley -> Allegra and Allegra -> Mary to the CHANGELOG. It's not as bad as I remembered in the end. Generators also extended :muscle:. It'd be good to start looking at code coverage at some point, because this is something I would have probably notice before...

clients/TypeScript/packages/client/README.md
```diff
@@ -0,0 +1,5 @@
+# Cardano Ogmios TypeScript Client Packages - Client
+
+Provides a natural interface for ECMAScript integrations, including abstraction of Websocket
+management and custom error mapping. The [schema](../schema/README.md) package is imported and
```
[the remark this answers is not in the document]
rhyslbw: Good point. I'll convert it to absolute so it will resolve in external registries such as npm.org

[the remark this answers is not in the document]
rhyslbw: https://github.com/KtorZ/cardano-ogmios/pull/29/commits/a5c1a4410c9cf52b15a3d31dca9efd74cfc036c3

clients/TypeScript/packages/client/src/Connection.ts
rhyslbw: https://github.com/KtorZ/cardano-ogmios/pull/29/commits/e234fd2ff8009f714a10a52f51af0abaf8d06acd

clients/TypeScript/packages/client/src/ChainSync/findIntersect.ts
rhyslbw: > I am not sure to get the rationale of not exposing the mirror/reflection to callers for findIntersect, but using it internally only

I decided to omit this capability with the local state queries too as I only see value in being able to mirror a value in `requestNext`.  Maybe I'm missing a use-case?

KtorZ: A use-case I can think of is using the reflection to drive a state-machine. I used to do this in some cases with Ogmios, where I would submit a state alongside any message and then, in the request handler, switch over that state to know what handler to call as a result. Arguably, one can use the response type directly for that to distinguish FindIntersect from RequestNext requests..

Another use-case can be for example, in the state query. One may send multiple `utxo` query for different set of addresses, and put the addresses in the mirror / reflection, so that the response handler has a way to figure out to which addresses correspond the UTxOs.

rhyslbw: > Arguably, one can use the response type directly for that to distinguish FindIntersect from RequestNext requests.
> ... One may send multiple utxo query for different set of addresses, and put the addresses in the mirror / reflection, so that the response handler has a way to figure out to which addresses correspond the UTxOs.

The values are returned as promises, so the context is captured in a variable.

https://github.com/KtorZ/cardano-ogmios/pull/29/files#diff-92254b7a19d36bcef42fdc1c0124b4150063bcee67b9dfdea248ffcf13ad072cR217-R220

``` Typescript
it('fetches the UTxO for the given addresses', async () => {
  const utxoSet = await utxo(['addr_test1qqymtheun4y437fa6cms4jmtfex39wzz7jfwggudwnqkdnr8udjk6d89dcjadt7tw6hmz0aeue2jzdpl2vnkz8wdk4fqz3y5m9'], { connection })
  expect(utxoSet[0]).toBeDefined()
})
```

rhyslbw: > This is also discarding messages that are not intersection responses which may accidentally lead to messages being skipped? 

The idea here is to just ignore messages we're not expecting, as all registered handlers will receive the message. I'll do some more concurrency testing to demonstrate this.

clients/TypeScript/packages/client/src/TxSubmission/errors.ts
rhyslbw: > I was wondering why this didn't cover Allegra and Mary errors

Ah yeah I should have raised an issue on this.

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
rhyslbw: Is there a chance of having it implemented at the protocol level, so that Ogmios stays aligned?

KtorZ: I don't see this happening :/ ...

KtorZ:
Rebased & fixed builds by adding the 3 tx submission errors newly introduced in Allegra / Mary.

PR #34, opened 2021-05-01
Setup Nix 

KtorZ:
Fixes #34 

```console
$ nix-build -A ogmios
```

Will build the project using Nix. Unfortunately, I couldn't reconciliate Nix with stack-based project ... but I am not willing to give up stack as a development tool. Maintaining both a cabal.project and stack.yaml will not be _too_ painful hopefully, but I've still added a Github Workflow to automatically build with Nix (and therefore, control that the cabal.project and stack.yaml remain somewhat in sync). 

Still to be done:

- [x] Better filter in `nix/pkgs.nix` to avoid putting in the store the entire project repository. `cleanGit` didn't work well with the package.yaml + hpack setup so I had to resort to using `cleanSourceWith` and custom filter. 

- [ ] Rework a bit the Makefile and build instruction to also mention building with Nix.

- [x] Get rid of the 'regenerate.sh' and whatnot which turned out to be unnecessary in the end.

- [ ] Investigate why Nix fails to pin the right git revision when building. 

- ~~[ ] Static linking would be great to have and would justify introducing Nix in the first place.~~

- ~~[ ] Maybe rework the Dockerfile build stage to build using Nix and create a final image from scratch using a static binary.~~


rhyslbw:
> Maybe rework the Docker image to build using Nix and create an image from scratch using a static binary.

I'm not in favour of this because it then requires nix to build the image from source.

KtorZ:
@rhyslbw I am not thinking about using nix on the host machine to produce a Dockerimage, but to use NixOs as a build stage to build a static binary as part of the image build.

Thus, only Docker is required to produce an image, the same way it is now.

rhyslbw:
> @rhyslbw I am not thinking about using nix on the host machine to produce a Dockerimage, but to use NixOs as a build stage to build a static binary as part of the image build.
> 
> Thus, only Docker is required to produce an image, the same way it is now.

Got it. I wasn't sure as you said rework the _Docker image_ rather than _Dockerfile_

KtorZ:
Yeah right. I don't want Nix to be a dependency on the host for whomever is building with Docker, so this would be a no-go for me. Worries not 😌

PR #37, opened 2021-05-04
Feature: ChainSyncClient.startSync

rhyslbw:
Adds a function to handle the sync setup, including request pipelining
to optimize performance. The default is 100 requests, but it's a configurable
argument. The `ChainSyncClient` no longer exposes `findIntersect`, and the awkward
`initialIntersection` logic removed.

- [x] Update REPL
- [x] Resolve open handle in the test suite
- [x] Fix CI workflow

### DIFF
--- a/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
+++ b/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
@@ -6,14 +6,12 @@ import {
   createClientContext
 } from '../Connection'
 import { UnknownResultError } from '../errors'
-import { baseRequest } from '../Request'
 import { createPointFromCurrentTip, ensureSocketIsOpen } from '../util'
 import { findIntersect, Intersection } from './findIntersect'
+import { requestNext } from './requestNext'
 
 export interface ChainSyncClient {
   context: InteractionContext
-  findIntersect: (points: Point[]) => ReturnType<typeof findIntersect>
-  initialIntersection: Intersection
   on: (messageHandlers: {
     rollBackward: (response: {
       point: Point,
@@ -28,6 +26,7 @@ export interface ChainSyncClient {
   }) => void
   requestNext: (options?: { mirror?: Mirror }) => void
   shutdown: () => Promise<void>
+  startSync: (points?: Point[], requestBuffer?: number) => Promise<Intersection>
 }
 
 export const createChainSyncClient = async (options?: {
@@ -38,23 +37,8 @@ export const createChainSyncClient = async (options?: {
       const { socket } = context
       socket.once('error', reject)
       socket.once('open', async () => {
-        const initialIntersection = await findIntersect(
-          [await createPointFromCurrentTip({ socket, closeOnCompletion: false })],
-          {
-            socket,
-            closeOnCompletion: false
-          }
-        )
         return resolve({
           context,
-          findIntersect: (points) => {
-            ensureSocketIsOpen(socket)
-            return findIntersect(points, {
-              socket,
-              closeOnCompletion: false
-            })
-          },
-          initialIntersection,
           on (messageHandlers) {
             socket.on('message', (message: string) => {
               const response: Ogmios['RequestNextResponse'] = JSON.parse(message)
@@ -79,17 +63,27 @@ export const createChainSyncClient = async (options?: {
           },
           requestNext (options) {
             ensureSocketIsOpen(socket)
-            socket.send(JSON.stringify({
-              ...baseRequest,
-              methodname: 'RequestNext',
-              mirror: options?.mirror
-            } as Ogmios['RequestNext']))
+            return requestNext(socket, options)
           },
           shutdown: () => new Promise(resolve => {
             ensureSocketIsOpen(socket)
             socket.once('close', resolve)
             socket.close()
-          })
+          }),
+          startSync: async (points, requestBuffer) => {
+            const intersection = await findIntersect(
+              points || [await createPointFromCurrentTip({ socket, closeOnCompletion: false })],
+              {
+                socket,
+                closeOnCompletion: false
+              }
+            )
+            ensureSocketIsOpen(socket)
+            for (let n = 0; n <= (requestBuffer || 100); n += 1) {
+              requestNext(socket)
+            }
+            return intersection
+          }
         })
       })
     }).catch(reject)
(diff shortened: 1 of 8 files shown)


KtorZ:
> The default is 1000 requests, 

I think `100` may be a more sensible default, according to [this table](https://ogmios.dev/mini-protocols/local-chain-sync/):

Screenshot from 2021-05-04 20-19-51

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -79,17 +63,27 @@ export const createChainSyncClient = async (options?: {
...
+          startSync: async (points, requestBuffer) => {
+            const intersection = await findIntersect(
+              points || [await createPointFromCurrentTip({ socket, closeOnCompletion: false })],
+              {
+                socket,
+                closeOnCompletion: false
+              }
+            )
+            ensureSocketIsOpen(socket)
+            for (let n = 0; n <= (requestBuffer || 1000); n += 1) {
```
KtorZ, line 82: either `let n = 1` or `n < (requestBuffer || 1000)`, otherwise it's off by one :s

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -79,17 +63,27 @@ export const createChainSyncClient = async (options?: {
...
-            } as Ogmios['RequestNext']))
+            return requestNext(socket, options)
           },
           shutdown: () => new Promise(resolve => {
             ensureSocketIsOpen(socket)
             socket.once('close', resolve)
             socket.close()
-          })
+          }),
+          startSync: async (points, requestBuffer) => {
```
KtorZ, line 73: Question: I am not sure how WebSockets handles event handlers subscription behind the scene but, should this function throw if called without any handlers defined for on 'rollForward' and on 'rollBackward' ?

KtorZ (commented):
:pray:

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
rhyslbw: I'll update it here too
https://github.com/KtorZ/cardano-ogmios/blob/47ac39dbbaf2869d27ba09e9361bab593775f0ff/server/static/tests/chain-sync.js#L84

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
rhyslbw: I'll move the handlers to a parameter of the upper function, as there's no longer a need to register the handlers after client construction.

rhyslbw: Done in 15ef6a79bef920e02e7fed4ff4e0efcb4535eacc

clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -11,76 +11,117 @@ import {
...
   })
-  it('selects the tip as the intersection if no point provided', async () => {
-    const client = await createChainSyncClient({ connection })
-    if (client.initialIntersection.point === 'origin' || client.initialIntersection.tip === 'origin') {
+
+  describe('startSync', () => {
+    it('selects the tip as the intersection if no point provided', async () => {
+      const client = await createChainSyncClient(stubHandlers, { connection })
+      const intersection = await client.startSync()
+      if (intersection.point === 'origin' || intersection.tip === 'origin') {
```
KtorZ, line 31: The second part of the assertion looks odd to me. I don't see why the node's tip would point to origin in this context?

clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -11,76 +11,117 @@ import {
...
+        })
+        await client.startSync(['origin'], requestBuffer)
+        await delay(800)
+        await client.shutdown()
+        expect(blocks.length).toBe(1000)
+        return 1000 * blocks.length / stop
+      }
+      const pipelinedBlocksPerSecond = await run()
+      const nonPipelinedBlocksPerSecond = await run(1)
+      expect(pipelinedBlocksPerSecond).toBeGreaterThan(nonPipelinedBlocksPerSecond)
```
KtorZ, line 117: Interesting 👌

server/static/tests/chain-sync.js
```diff
@@ -81,7 +81,7 @@ const lastAllegraBlock = {
             default:
               if (!didBootstrap) {
                 didBootstrap = true;
-                for(let n = 0; n <= N_BOOTSTRAP; n += 1) {
+                for(let n = 0; n < N_BOOTSTRAP; n += 1) {
```
KtorZ, line 84: Thank you.

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -6,90 +6,85 @@ import {
...
+              reflection: response.reflection
+            })
+          } else if ('RollForward' in response.result) {
+            messageHandlers.rollForward({
+              block: response.result.RollForward.block,
+              tip: response.result.RollForward.tip,
+              reflection: response.reflection
+            })
+          } else {
+            throw new UnknownResultError(response.result)
```
KtorZ, line 58: I am still a bit doubtful about this w.r.t. concurrency. I assume this is fine if a new connection is used for the state queries, but if re-used, then this may cause issues? I should check back how the state query client is implemented maybe to dispel my doubts.

KtorZ (approved):
✅

clients/TypeScript/packages/client/test/ChainSync.test.ts
rhyslbw: It's a guard to ensure the chain tip has progressed past the `origin` point. If for some reason this is the case, the current state would produce the same result as the following test case `intersects at the genesis if origin provided as point`.

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
rhyslbw: Each client has an isolated connection, however, the scenario you're describing is possible by passing in an `InteractionContext` so is worth documenting with a test.

rhyslbw: https://github.com/KtorZ/cardano-ogmios/issues/40

PR #47, opened 2021-05-10
chore: add release workflow

rhyslbw:
Currently the only job is to publish the TS client packages,
however this could evolve into a release workflow that automates
all tasks, triggered manually. NPM_TOKEN is required, NPM_REGISTRY
is optional.

- [x] Sync version with repository (3.2.0) - > Moved to https://github.com/KtorZ/cardano-ogmios/pull/48
- [x] Build and attach static packaged REPL
- [x] Add documentation for installing via npm.org -> Moved to https://github.com/KtorZ/cardano-ogmios/pull/48


.github/workflows/post-release.yaml
```diff
@@ -0,0 +1,51 @@
...
+          registry: ${{ secrets.NPM_REGISTRY }}
+          token: ${{ secrets.NPM_TOKEN }}
+
+      - name: 📤 Publish schema package to npm registry
+        uses: JS-DevTools/npm-publish@v1
+        with:
+          check-version: true
+          package: clients/TypeScript/packages/schema/package.json
+          registry: ${{ secrets.NPM_REGISTRY }}
+          token: ${{ secrets.NPM_TOKEN }}
```
KtorZ, line 51: :+1:

clients/TypeScript/README.md
```diff
@@ -1,9 +1,9 @@
 # Cardano Ogmios TypeScript Client Packages
-A Yarn Workspace containing three packages:
+A Yarn Workspace containing three packages, each published to the npm.org registry:
 
-- [@cardano-ogmios/client](./packages/client/README.md)
-- [@cardano-ogmios/schema](./packages/schema/README.md)
-- [@cardano-ogmios/repl](./packages/repl/README.md)
+- [@cardano-ogmios/client](./packages/client)
+- [@cardano-ogmios/schema](./packages/schema)
+- [@cardano-ogmios/repl](./packages/repl)
```
KtorZ, line 6: :clap:

clients/TypeScript/packages/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cardano-ogmios/client",
-  "version": "0.1.2",
+  "version": "3.2.0",
```
KtorZ, line 3: I wanted to propose that some time ago but then, thought it may be a good idea to keep the clients and the server on their own versioning. There may be new versions of the client(s) without any new changes from the server :thinking:. Unless we treat everything in the repository (client(s) + server) as _one_ "ogmios" entity and everything is released at the same time. 

Incidentally, I asked myself this question again yesterday while finalizing the changelog for the release. There's currently no changelog for any of the client package. Here again, if decide to move all modules in lockstep then I'd also be in favor of keeping _one_ global changelog and have symlinks from each client package root to that global changelog.

.github/workflows/post-release.yaml
KtorZ: (You have the necessary rights to configure the secrets on the repository, right ?)

clients/TypeScript/packages/client/package.json
rhyslbw: > may be a good idea to keep the clients and the server on their own versioning.

That was also my initial thought, but considering we're in the same code repository it makes sense to release at the same time. If we need separation, then lifting the client to a new repo (under and org space) is the way to go IMO.

rhyslbw: Having said that, I do feel a single repo has a number of advantages:

1. The schema package, which is fully derived from the JSON-WSP schema, is then simultaneously available with server releases.
2. Ogmios REPL is the universal quick start UI, not just a TypeScript-focused product.
3. The client is used to provide integration test coverage.

KtorZ: I agree with all your points above :+1: 
Would you mind doing --> 

> have symlinks from each client package root to that global changelog. 

Pretty please :pray:

rhyslbw:
I will rework this PR to be a _release_ workflow, as I think we can tie everything into a nice process.

KtorZ:
Added `NPM_TOKEN` to Github secrets :+1:

KtorZ (approved):

PR #58, opened 2021-05-15
Rework Dockerfile in multi-stages and provide a stage where cardano-node & ogmios are bundled together

### LINKED ISSUE #44: Bundle `cardano-node` into the Ogmios Docker image
Build `cardano-node` and `cardano-ogmios` in the same container instead of running two Docker containers and adopting the strategy of a [shared volume](https://github.com/KtorZ/cardano-ogmios/blob/master/docker-compose.yml#L36) for the domain socket.

#### Why is it a good idea?
As `cardano-ogmios` is very lightweight and simply expresses the underlying protocols, it's a natural idea to think about it as a type of node that can be communicated with over TCP. It would eliminate a common source of issues with the docker volume approach, including custom deployments not using the provided `docker-compose.yaml`, and reduce the deployment barrier to entry to a single `docker run` command.
 
No

KtorZ:
Fixes #44 

---

- 4fea04a3986568f7e3d04fc69394597359bd4aa7
  :round_pushpin: **Rework Dockerfile in multi-stages and provide a stage where cardano-node & ogmios are bundled together**
    One can still fallback to the previous behavior by running 'docker build . --target ogmios'; The full build now however also builds a cardano-node.

- 362bbddfaf59ccdb67f9d55217cf61078a937bf1
  :round_pushpin: **Adjust Dockerfile to use cabal only, and pre-build Ogmios during the setup.**
  
- b98ba183cf2b762b478aaf52464f8cb637a05376
  :round_pushpin: **Fix ARG / scheme for cardano-node-ogmios build stage**
  
- ac06142310965f82265832c0e146f2631d5d01ba
  :round_pushpin: **Add more debug capabilities to wait-for-sync script.**
  
- 96f1ad8cfa7d96c67a8d46a930f589cf7d89f314
  :round_pushpin: **Fix Dockerlint warnings & errors.**
  
- 69401b5c3a2f5334cddc3c6cb45a446ebfbfc195
  :round_pushpin: **Add workflow to build and push Docker images using github actions (instead of docker hub)**
  
- c4838bc242d2e18a51df3d4f60c734e8e87e4a0e
  :round_pushpin: **Include instructions about building/running standalone cardano-node-ogmios in the user-guide.**


rhyslbw (commented):
> Not sure we can restrict Dockerhub builds to a specific build target so we might as well disable builds in Dockerhub and do everything with Github actions.

Looks like you can do it with [custom build hooks](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks)

KtorZ:
So, this now work fine: 

```console
$ docker build --target ogmios .
$ docker build --target cardano-ogmios-node .
```

In the case of the latter, one can then:

```console
$ docker run --rm --name cardano-node-ogmios -e NETWORK=testnet -p 1337:1337 ...
```

which starts Ogmios + a node. Works with mainnet, testnet and staging aaaaand, seems like the node is syncing fine (after struggling a good bit, finally found out that the runtime image needs to have `netbase` installed without what the node fails to subscribe to its peers). 

I still need to adjust the user manual and maybe some bits in the CI accordingly but, it's mostly there :whale:

docs/content/getting-started/docker.md
```diff
@@ -48,3 +48,17 @@ For example, for running cardano-node + ogmios on the testnet, listening to tcp/
 ```console
 $ NETWORK=testnet OGMIOS_PORT=1338 docker-compose --project-name cardano-ogmios-testnet up
 ```
+
+## As a single image
```
rhyslbw, line 52: As this is the default when building the image, maybe it's best to lead with this as it is certainly a faster way to get up and running.

docs/content/getting-started/docker.md
```diff
@@ -48,3 +48,17 @@ For example, for running cardano-node + ogmios on the testnet, listening to tcp/
...
+
+If you're not fond of docker-compose and the container orchestration it requires, you may also build (or pull) an image to run both a `cardano-node` and `ogmios` side-by-side, within the same container. Note that this is the default build target when building ogmios with Docker, but we can be explicit when building the image using `--target cardano-node-ogmios`. For example:
```
rhyslbw, line 54: If this is positioned first...
```suggestion
The default Ogmios Docker image comes with a `cardano-node` relay, and is the easiest way to get up an running.
```

docs/content/getting-started/docker.md
```diff
@@ -48,3 +48,17 @@ For example, for running cardano-node + ogmios on the testnet, listening to tcp/
...
+
+```console
+$ docker build --target cardano-node-ogmios --tag ktorz/cardano-node-ogmios:latest .
```
rhyslbw, line 57:
```suggestion
$ docker build --tag ktorz/cardano-node-ogmios:latest . 
```

docs/content/getting-started/docker.md
```diff
@@ -48,3 +48,17 @@ For example, for running cardano-node + ogmios on the testnet, listening to tcp/
...
 ```
+
+## As a single image
+
+If you're not fond of docker-compose and the container orchestration it requires, you may also build (or pull) an image to run both a `cardano-node` and `ogmios` side-by-side, within the same container. Note that this is the default build target when building ogmios with Docker, but we can be explicit when building the image using `--target cardano-node-ogmios`. For example:
+
+```console
+$ docker build --target cardano-node-ogmios --tag ktorz/cardano-node-ogmios:latest . 
+```
+
```
rhyslbw, line 59: _Note that this is the default build target for the image, but `--target cardano-node-ogmios` can be set explicitly_

docs/content/getting-started/docker.md
rhyslbw: Think it's best to use no target here, then add a note below

docs/content/getting-started/docker.md
```diff
@@ -27,10 +27,10 @@ $ docker-compose up
 To build the Ogmios image from sources, pass the `--build` flag. This is useful if you need a different version than the latest one available on Dockerhub.  
 
 {{% notice tip %}}
-If you're building locally using `docker build`, make sure to leverage existing cache steps from Docker Hub setting the env `DOCKER_BUILDKIT=1` and passing `--cache-from ktorz/ogmios:latest`. For `docker-compose build` or `docker-compose up --build`, **also** set`COMPOSE_DOCKER_CLI_BUILD=1`. A full build of Ogmios without cache may take up to 45 minutes!
+If you're building locally using `docker build --target ogmios`, make sure to leverage existing cache steps from Docker Hub setting the env `DOCKER_BUILDKIT=1` and passing `--cache-from ktorz/ogmios:latest`. For `docker-compose build` or `docker-compose up --build`, **also** set`COMPOSE_DOCKER_CLI_BUILD=1`. A full build of Ogmios without cache may take up to 45 minutes!
```
rhyslbw, line 30: The cache is working for me when not targeting `ogmios`. Is there a reason you're scoping this?

docs/content/getting-started/docker.md
rhyslbw: Ideally though, using a remote build context means a consumer can run a build command without cloning the repo first. 

This is [valid](https://docs.docker.com/engine/reference/commandline/build/#git-repositories):
```
$ DOCKER_BUILDKIT=1 docker build --cache-from ktorz/ogmios:latest --tag ktorz/cardano-node-ogmios:latest https://github.com/KtorZ/cardano-ogmios.git#:server
```

... however [Buildkit doesn't support sub-directories](https://github.com/moby/buildkit/issues/1684) :-1: 

Considering the Docker image now represents more than just the Ogmios Server, it does hint at hoisting it to the top level. This is similar to how Cardano Rosetta was implemented: https://github.com/input-output-hk/cardano-rosetta. Is that an option?

docs/content/getting-started/docker.md
```diff
@@ -48,3 +48,17 @@ For example, for running cardano-node + ogmios on the testnet, listening to tcp/
...
+If you're not fond of docker-compose and the container orchestration it requires, you may also build (or pull) an image to run both a `cardano-node` and `ogmios` side-by-side, within the same container. Note that this is the default build target when building ogmios with Docker, but we can be explicit when building the image using `--target cardano-node-ogmios`. For example:
+
+```console
+$ docker build --target cardano-node-ogmios --tag ktorz/cardano-node-ogmios:latest . 
+```
+
+Images with either release tags or `:latest` tags are also uploaded to the corresponding [DockerHub registry](https://hub.docker.com/r/ktorz/cardano-node-ogmios) and you may also pull them directly from there. Assuming you've pulled `ktorz/cardano-node-ogmios:latest`, you can start a container by running:
+
+```console
+$ docker run --detach --name cardano-node-ogmios --e NETWORK testnet -p 1337:1337 ktorz/cardano-node-ogmios:latest
```
rhyslbw, line 63: I've been using `1338` for `testnet`. Do you think it's worth using this convention to help with support and education?

```
-p 1338:1337
```

rhyslbw: Also, in quick start guides I feel it's better to provide the run command without specifying detached mode, as the goal is to just see it working. Another option is to add `&& docker logs -f  cardano-node-ogmios`, but I prefer the former since anyone needing the process to be detached probably knows what to do.

docs/content/getting-started/docker.md
KtorZ: I've been trying to keep the first part of the docker doc about "standalone ogmios", while the second part is more about building/running "cardano-node-ogmios". Obviously, this ain't clear enough :sweat_smile: ... The entire page should probably be re-written now since this PR changes quite a lot what we can do with docker.

docs/content/getting-started/docker.md
KtorZ: > I've been using 1338 for testnet. Do you think it's worth using this convention to help with support and education?

And I've been wondering why :sweat_smile: ? I see little chance for people running both mainnet and testnet at the same time, and network discriminant in addresses will already provide a good mechanism for silly mistakes. 

I fail to see the rationale here :no_mouth:  ?

---

1. :heavy_check_mark: 
2. :heavy_check_mark:

docs/content/getting-started/docker.md
KtorZ: Awe :( ... Well, it's also possible to move the `Dockerfile` one level up indeed, though I like it well scoped within `server`. The compose file is already at the root level however, and building it without cloning the repository sounds like an argument convincing enough to move it there :+1:

docs/content/getting-started/docker.md
rhyslbw: > I fail to see the rationale here no_mouth ?

If you bind `1337` to a testnet container, then you have to destroy it before running a `mainnet` container on the same host. Not a big deal, so happy to stick with `1337`

PR #63, opened 2021-05-20
fix: client TypeScript chain-sync sequential processing

### LINKED ISSUE #61: Chain-sync client pipelining flaw
#### What Git revision are you using? 
`d76837292fd8c3aa2edd4f2b363bab4aa8cded38`

#### What operating system are you using, and which version?

- [x] Linux / Ubuntu
- [ ] Linux / Other
- [ ] OSX
- [ ] Windows

#### Describe what the problem is? 
The current design makes n `requestNext` messages to prime the pipeline, however the first rollback operation should be performed _**before**_ handling the first `rollForward` instruction as there may be state to cleanup. 

https://github.com/KtorZ/cardano-ogmios/blob/master/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts#L74-L87

#### What should be the expected behavior?
- Priming the pipeline should take place after the first `requestNext` request is sent from the implementation.
- The implementor should also not be concerned with it being the initial request or subsequent requests.

rhyslbw:
**feature: client TypeScript chain-sync client requestNext improvement**

**BREAKING CHANGE:** ChainSyncClient no longer exposes a `requestNext`
function. Instead you must invoke the callback provided as the second
argument in each of rollBackward and rollForward handlers.

**BREAKING CHANGE:** ChainSyncClient no longer exposes JSON-WSP reflection,
as there would be unexpected results given the first n messages would all
share the same reflected value.
_____

**fix: queue client TypeScript chain-sync requestNext responses**
Implements an in-memory queue to ensure `requestNext` responses are processed
sequentially when there are async operations in the message handlers.

**BREAKING CHANGE:** The `ChainSyncClientMessageHandlers` methods now must return
a promise.

Note: The dependency introduced here is https://www.npmjs.com/package/fastq.

### DIFF
--- a/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
+++ b/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
@@ -15,7 +15,7 @@ export interface ChainSyncClient {
   shutdown: () => Promise<void>
   startSync: (
     points?: Point[],
-    requestBuffer?: number
+    inFlight?: number
   ) => Promise<Intersection>
 }
 
@@ -78,7 +78,7 @@ export const createChainSyncClient = async (
             socket.once('close', resolve)
             socket.close()
           }),
-          startSync: async (points, requestBuffer) => {
+          startSync: async (points, inFlight) => {
             const intersection = await findIntersect(
               points || [await createPointFromCurrentTip({ socket, closeOnCompletion: false })],
               {
@@ -87,7 +87,7 @@ export const createChainSyncClient = async (
               }
             )
             ensureSocketIsOpen(socket)
-            for (let n = 0; n < (requestBuffer || 100); n += 1) {
+            for (let n = 0; n < (inFlight || 100); n += 1) {
               requestNext(socket)
             }
             return intersection
--- a/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
+++ b/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
@@ -1,7 +1,6 @@
 import { Block, Ogmios, Point, Tip } from '@cardano-ogmios/schema'
 import {
   ConnectionConfig,
-  Mirror,
   InteractionContext,
   createClientContext
 } from '../Connection'
@@ -12,29 +11,35 @@ import { requestNext } from './requestNext'
 
 export interface ChainSyncClient {
   context: InteractionContext
-  requestNext: (options?: { mirror?: Mirror }) => void
   shutdown: () => Promise<void>
-  startSync: (points?: Point[], requestBuffer?: number) => Promise<Intersection>
+  startSync: (
+    points?: Point[],
+    requestBuffer?: number
+  ) => Promise<Intersection>
 }
 
 export interface ChainSyncMessageHandlers {
-  rollBackward: (response: {
-    point: Point,
-    tip: Tip,
-    reflection: Mirror
-  }) => void
-  rollForward: (response: {
-    block: Block,
-    tip: Tip,
-    reflection: Mirror
-  }) => void
+  rollBackward: (
+    response: {
+      point: Point,
+      tip: Tip
+    },
+    requestNext: () => void
+  ) => void
+  rollForward: (
+    response: {
+      block: Block,
+      tip: Tip
+    },
+    requestNext: () => void
+  ) => void
 }
 
 export const createChainSyncClient = async (
   messageHandlers: ChainSyncMessageHandlers,
   options?: {
-  connection?: ConnectionConfig
-}): Promise<ChainSyncClient> => {
+    connection?: ConnectionConfig
+  }): Promise<ChainSyncClient> => {
   return new Promise((resolve, reject) => {
     createClientContext(options).then(context => {
       const { socket } = context
@@ -45,15 +50,13 @@ export const createChainSyncClient = async (
           if ('RollBackward' in response.result) {
             messageHandlers.rollBackward({
               point: response.result.RollBackward.point,
-              tip: response.result.RollBackward.tip,
-              reflection: response.reflection
-            })
+              tip: response.result.RollBackward.tip
+            }, () => requestNext(socket))
           } else if ('RollForward' in response.result) {
             messageHandlers.rollForward({
               block: response.result.RollForward.block,
-              tip: response.result.RollForward.tip,
-              reflection: response.reflection
-            })
+              tip: response.result.RollForward.tip
+            }, () => requestNext(socket))
           } else {
             throw new UnknownResultError(response.result)
           }
@@ -62,10 +65,6 @@ export const createChainSyncClient = async (
       socket.once('open', async () => {
         return resolve({
           context,
-          requestNext (options) {
-            ensureSocketIsOpen(socket)
-            return requestNext(socket, options)
-          },
           shutdown: () => new Promise(resolve => {
             ensureSocketIsOpen(socket)
             socket.once('close', resolve)
--- a/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
+++ b/clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
@@ -5,6 +5,7 @@ import {
   createClientContext
 } from '../Connection'
 import { UnknownResultError } from '../errors'
+import fastq from 'fastq'
 import { createPointFromCurrentTip, ensureSocketIsOpen } from '../util'
 import { findIntersect, Intersection } from './findIntersect'
 import { requestNext } from './requestNext'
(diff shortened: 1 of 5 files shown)


clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -46,24 +50,25 @@ describe('ChainSync', () => {
...
           if (blocks.length < 10) {
             blocks.push(block)
-            client.requestNext()
           }
+          requestNext()
         }
       }, {
-        connection
+        connection,
+        requestBuffer: 10
```
KtorZ, line 69: Request buffer is perhaps a bit too vague to convey the right meaning? Especially in the context of networking where many things can be buffered. This refers to the number of requests in-flight, so perhaps simply `inFlight`?

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -12,29 +11,33 @@ import { requestNext } from './requestNext'
 
 export interface ChainSyncClient {
   context: InteractionContext
-  requestNext: (options?: { mirror?: Mirror }) => void
```
KtorZ, line 15: So this means that the chain-sync is necessarily pipelined?

KtorZ: Actually no, because one can always specify a `requestBuffer: 1` to get a one-by-one behavior (if that's even a useful use-case :thinking: ?)

KtorZ (approved):
Looks good.

rhyslbw:
I need to rework this as the following test is a better specification of the correct behaviour, however is currently failing:

``` TypeScript
it('processes messages sequentially in order', async () => {
    const log: string[] = []
    const randomMsUpTo = (max: number) => (Math.random() * (max - 5 + 1) ) << 0
    const client = await createChainSyncClient({
      rollBackward: async (_response, requestNext) => {
        log.push('rollBackward received')
        // Simulate some work being done
        await delay(randomMsUpTo(50))
        log.push('rollBackward processed')
        requestNext()
      },
      rollForward: async (_response, requestNext) => {
        log.push('rollForward received')
        // Simulate some work being done
        await delay(randomMsUpTo(25))
        log.push('rollForward processed')
        requestNext()
      }
    }, {
      connection,
      requestBuffer: 3
    })
    await client.startSync(['origin'])
    await delay(500)
    await client.shutdown()
    expect(log.slice(0,8)).toStrictEqual([
      'rollBackward received',
      'rollBackward processed',
      'rollForward received',
      'rollForward processed',
      'rollForward received',
      'rollForward processed',
      'rollForward received',
      'rollForward processed'
    ])
  })
  ```
  
  ``` console
   FAIL  test/ChainSync.test.ts
  ● ChainSync › processes messages sequentially in order

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Array [
        "rollBackward received",
        "rollBackward processed",
        "rollForward received",
    -   "rollForward processed",
        "rollForward received",
    -   "rollForward processed",
        "rollForward received",
    +   "rollForward processed",
        "rollForward processed",
    +   "rollForward received",
      ]

      151 |     await delay(500)
      152 |     await client.shutdown()
    > 153 |     expect(log.slice(0,8)).toStrictEqual([
          |                            ^
      154 |       'rollBackward received',
      155 |       'rollBackward processed',
      156 |       'rollForward received',

      at Object.<anonymous> (ChainSync.test.ts:153:28)
```

KtorZ:
@rhyslbw I don't see what's correct in the behavior from the test scenario above :sweat_smile:  ? With the delays, I'd expect the event loop to pass the thread execution to a different event callback, and it sounds to me that a sequence like:

```
      'rollBackward received',
      'rollBackward processed',
      'rollForward received',
      'rollForward received',
      'rollForward received',
      'rollForward processed',
      'rollForward processed',
      'rollForward processed'
```

is more what I'd expect. After the rollback is processed, the `requestNext` now immediately put a 100 request messages in flight. And because of the delays (which, I assume uses `setTimeout` behind the scene), the thread preemption is given away to other threads. Depending on the delay, I can totally see 3 messages to have already been received before any of them is processed! 

To be frank, I don't know how to reliably simulate a busy-loop in Node.js because any timeout will give priority to a new callback of the event loop and, any fake `for loop` for instance would get removed by the JIT compiler if they don't produce any meaningful work for the rest of the program.

Note that is most likely NOT what one wants in such callbacks. Likely, one would work around that by having an artificial queue that each callback can write into, and then make sure that only one worker processes elements from that queue.

clients/TypeScript/packages/client/test/ChainSync.test.ts
rhyslbw: 5ddae7e9054f3d8ebd2159915990ce8088609c03

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -1,77 +1,84 @@
...
   messageHandlers: ChainSyncMessageHandlers,
   options?: {
-  connection?: ConnectionConfig
-}): Promise<ChainSyncClient> => {
+    connection?: ConnectionConfig
+  }): Promise<ChainSyncClient> => {
   return new Promise((resolve, reject) => {
     createClientContext(options).then(context => {
       const { socket } = context
+      const queue = fastq.promise(async (response) => {
```
rhyslbw, line 47: Sequential queue introduced to control execution flow

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -1,77 +1,84 @@
...
-          } else if ('RollForward' in response.result) {
-            messageHandlers.rollForward({
-              block: response.result.RollForward.block,
-              tip: response.result.RollForward.tip,
-              reflection: response.reflection
-            })
-          } else {
-            throw new UnknownResultError(response.result)
-          }
+          queue.push(response)
```
rhyslbw, line 70: Response pushed onto queue

clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -116,12 +126,55 @@ describe('ChainSync', () => {
       const nonPipelinedBlocksPerSecond = await run(1)
       expect(pipelinedBlocksPerSecond).toBeGreaterThan(nonPipelinedBlocksPerSecond)
     })
+
+    it('processes messages sequentially in order', async () => {
```
rhyslbw, line 130: New test

rhyslbw:
Thanks @KtorZ , yes I should have been more clear with my post. It was more to give you the heads-up on the solution pivot by providing a test case that, for me at least :smile: , documented the expected behaviour of being able to use async operations within the callback handlers. The new solution actually highlights the original issue was not well defined, so I've removed the change to offset sending `n` `requestNext`  messages, but have retained the new callback approach as it's a nicer API.

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -62,12 +63,15 @@ export const createChainSyncClient = async (
         } else {
           throw new UnknownResultError(response.result)
         }
-      }, 1)
+      }
+      const responseHandler = options?.sequential !== false ?
```
rhyslbw, line 67: The response handler is set to the underlying message handler function if configured for non-sequential processing.

clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -127,50 +163,30 @@ describe('ChainSync', () => {
...
-        'rollForward received: Block 4',
-        'rollForward processed: Block 4'
-      ])
+      expect(resultLog.slice(0, 12)).toStrictEqual(sequentialResponses)
     })
   })
 
+  it('can be configured to processes messages as fast as possible, when sequential processing is not required', async () => {
+    const resultLog: string[] = []
+    const client = await createChainSyncClient(messageOrderTestHandlers(resultLog), {
```
rhyslbw, line 180: New test

clients/TypeScript/packages/client/src/ChainSync/ChainSyncClient.ts
```diff
@@ -1,77 +1,88 @@
...
+            tip: response.result.RollForward.tip
+          }, () => {
+            requestNext(socket)
+          })
+        } else {
+          throw new UnknownResultError(response.result)
+        }
+      }
+      const responseHandler = options?.sequential !== false
+        ? fastq.promise(messageHandler, 1).push
```
KtorZ, line 68: That is one strange API, got me confused at first and had to look it up after looking for a corresponding `pop` in vain :sweat_smile: ...

clients/TypeScript/packages/client/test/ChainSync.test.ts
```diff
@@ -116,12 +162,35 @@ describe('ChainSync', () => {
...
+      connection,
+      sequential: false
+    })
+    await client.startSync(['origin'], 3)
+    await delay(500)
+    await client.shutdown()
+    expect(resultLog.slice(0, 12)).not.toStrictEqual(sequentialResponses)
```
KtorZ, line 187: :+1:

KtorZ (approved):

PR #66, opened 2021-05-29
Alonzo - Part 2 (JSON Schema for Alonzo data-types) 

KtorZ:
- :round_pushpin: **Cover Alonzo[Block] in the JSON-WSP specification.**
    This does not yet cover the new errors coming from the transaction submission. So I don't expect the corresponding tests to pass, but the RequestNextResponse now should when enabling Alonzo generators.

- :round_pushpin: **Generate AlonzoBlock as part of the JSON test suite**
    Weirdly enough, I couldn't resort to a simple 'Arbitrary' this time, because it seems that the transactions within blocks are different from the transactions generated by the arbitrary instance. So, I had to fmap on the sequence of Tx using a newly introduced 'toTxSeq'. I don't know if the tests pass since I am running some heavy simulations on the side and my laptop will just die of exhaustion if I run the test ... I knew 16GB of RAM was not enough. Anyway, CI will tell.

- :round_pushpin: **Wrap up JSON-WSP definitions for Alonzo blocks. Tests now pass.**
  
- :round_pushpin: **Return vkey signatures as a map instead of a list of singletons.**
  
- :round_pushpin: **Make JSON tests a bit faster by reading schema from disk only once per test group.**
    Instead of once for every prop execution! From 80s to 55s. Not much, but that's good to take :).

- :round_pushpin: **Start documenting new Alonzo SubmitTx errors in JSON-WSP**
  
- :round_pushpin: **Wrap up JSON-WSP definitions for Alonzo types.**
  
- :round_pushpin: **Merge TxOut JSON definitions together**
    Instead of carrying around 3 definitions for TxOut in Shelley/Allegra, TxOut in Mary and TxOut in Alonzo. I've also changed the base definition of 'value' in Pre-Mary eras to be a singleton object with a single field 'coins', so that it's uniform / compatible with next eras. This is however yet-another-breaking change.

- :round_pushpin: **Add missing Alonzo generators for state-queries scenarios.**
  
- :round_pushpin: **Rework a bit error reporting from hspec-json-schema to be less noisy / more helpful**
    The Ogmios schema is more than 4000 lines long, with a lot of models and nested schemas, so spitting schemas back on stdout for each errors can get very verbose. It started from a good intention,but turns out to be quite useless in practice. In the end, having the failing JSON and the errors is more useful for debugging.

- :round_pushpin: **Make sure to also build modules' tests files in the continuous integration workflow**
    The tests aren't executed in CI (only the server tests are). Yet, we should still build them to make sure everything in the source at least compiles.

- :round_pushpin: **Add support for testing with Golden files + fix some tricky schema issues found by chance**
    Sometimes, it's hard to reproduce failing cases with tests and debugging can be quite hard. Now, one can add a golden JSON file (obtained from the hspec-json-schema error output) to be expicitely tested, like a pre-generated case. Currently versioning some of tricky cases which popped up sporadically while testing other things.

- :round_pushpin: **More CHANGELOG update related to Alonzo.**
  
- :round_pushpin: **Update clients/TypeScript**
    I ended up fixing a few things in the JSON schema which has been bothering me for a while. Since the next update will be a major upgrade, it makes sense to pack breaking changes into it. I've therefore tweaked the JSON-WSP specifications to void generating types with identical names, forcing the generating to distinguish them using indexes (e.g. TxTooLarge vs TxTooLarge1). This commit also adds new tx submission errors and fixes the Utxo type now merged into one.

### DIFF
--- a/clients/TypeScript/packages/client/src/StateQuery/queries/utxo.ts
+++ b/clients/TypeScript/packages/client/src/StateQuery/queries/utxo.ts
@@ -2,9 +2,9 @@ import {
   Address,
   EraMismatch,
   Ogmios,
-  Utxo1,
-  Utxo2,
-  UtxoMary
+  TxIn,
+  TxOut,
+  Utxo
 } from '@cardano-ogmios/schema'
 import { EraMismatchError, QueryUnavailableInCurrentEraError, UnknownResultError } from '../../errors'
 import { InteractionContext } from '../../Connection'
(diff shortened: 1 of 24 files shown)


rhyslbw:
> I ended up fixing a few things in the JSON schema which has been bothering me for a while. 
:raised_hands: 

I actually drafted an issue about this, but it was lost and I clearly forgot to redo it. Thanks for proactively fixing

clients/TypeScript/packages/client/src/StateQuery/queries/utxo.ts
```diff
@@ -2,9 +2,9 @@ import {
   Address,
   EraMismatch,
   Ogmios,
-  Utxo1,
-  Utxo2,
-  UtxoMary
+  TxIn,
+  TxOut,
+  Utxo,
```
rhyslbw, line 7: `yarn lint --fix`

rhyslbw (changes requested):
LGTM, just need to fix the lint failure

PR #74, opened 2021-06-10
Unit tests for Ouroboros mini-protocols using simulated IO

KtorZ:
- :round_pushpin: **Write unit test using io-sim for the ChainSync protocol.**
    For now, it only covers basic scenario but, the instrumentation behind the scene is pretty heavy. In particular, it could (would) be used to generate a random sequence of input chain sync messages, for then we can ensure some properties like: order preservation, expected response, no deadlock / errors etc... It'd be interesting to also do it for the local state query and the health check, to make sure that all cases are properly handled by the logic. Everything runs in IO-sim, so it's blazing fast and doesn't require any real networking setup :tada:

- :round_pushpin: **Make max in flight (a.k.a. pipelined) chain-sync requests configurable**
    This is mainly for enabling unit test scenarios which saturate the queue without having to generate thousands of events.

- :round_pushpin: **Generate arbitrary mirror values of chain-sync spec instead of hard-coding them.**
  
- :round_pushpin: **Allow WSP handlers to reply with a fault and preserve reflection value**
  
- :round_pushpin: **Add property for checking interleaves of requestNext and findIntersect**
  
- :round_pushpin: **Remove redundant code identified with property tests.**
  
- :round_pushpin: **Lower down a bit QC confidence for faster chain-sync unit tests**
    The property are really generating two cases (Nothing or Just _), the value in Just can really be anything, so in the end, running the property a 100 times is _overkill_

- :round_pushpin: **Also cover the state-query client through property testing in IO-sim**

TODO: 

- [ ] Test the health check and tip sync with the same approach

### DIFF
--- a/server/modules/cardano-client/src/Cardano/Network/Protocol/NodeToClient.hs
+++ b/server/modules/cardano-client/src/Cardano/Network/Protocol/NodeToClient.hs
@@ -24,6 +24,17 @@ module Cardano.Network.Protocol.NodeToClient
     -- * Connecting
     , connectClient
     , codecs
+    , cChainSyncCodec
+    , cTxSubmissionCodec
+    , cStateQueryCodec
+    , nodeToClientV_Latest
+
+    -- * Running
+    , runPeer
+    , runPipelinedPeer
+    , chainSyncClientPeerPipelined
+    , localStateQueryClientPeer
+    , localTxSubmissionClientPeer
 
     -- * Helpers / Re-exports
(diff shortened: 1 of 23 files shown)


rhyslbw:
Sounds like this could be the foundation for an Ogmios integration test mode :thinking:

KtorZ:
Hmm. I didn't thought on it that way but yes. There's maybe something to be done in that direction 🤔

PR #76, opened 2021-06-15
Refactor: config ownership and build-time network selection

rhyslbw:
_This is a proposal to change the config strategy in both `cardano-node-ogmios` and `ogmios` itself._
________________________________________________________________________________

**Context**
It's currently not possible to run an instance of `cardano-node-ogmios` on a network other than `mainnet` or `testnet`, including a custom network such as a local network being run for the purpose of integration testing. 

This refactor takes ownership of `cardano-node` configuration, which was previously
 being downloaded from the IOHK config registry during the `cardano-node-ogmios` build
 and then manipulated to fit the internal requirements. This approach places a dependency
 on the centralised service, which is a risk and brittle when wishing to modify the values
 more extensively.

 This design changes the network selection to a build-time action, so that containers
 are provided with a single config dir at runtime, but depends on a change in Ogmios
 to be provided with the same config to be able to read information from the genesis files.
 
 This pattern was used when implementing `cardano-rosetta`, and was [discussed in the Rosetta community forum](https://community.rosetta-api.org/t/multiple-networks/167/3?u=rhysbw) where more context is available.
 
 **Todo:**
 - [x] - Refactor Ogmios server to accept `--config` option, which provides the mapping to the genesis files via the `cardano-node` config file, in place of using `OGMIOS_NETWORK`.
 - [x] - Include a matrix of networks in the Github Workflow to build and tag for Docker Hub.


server/src/Ogmios/App/Options.hs
```diff
@@ -240,92 +241,48 @@ deriving newtype instance ToJSON EpochSlots
...
--- | Lookup environment for a given version data name, default to mainnet.
-lookupNetworkParameters
-    :: IO NetworkParameters
-lookupNetworkParameters = do
-    mStr <- lookupEnv envOgmiosNetwork
-    let params = maybe (Just mainnetNetworkParameters) parseNetworkParameters mStr
-    maybe (die err) pure params
   where
-    err = "Couldn't parse " <> envOgmiosNetwork <> ". Have a look at the usage using '--help'."
-
```
rhyslbw, line 255: Nice simplification here :sunglasses:

PR #77, opened 2021-06-18
Refactor/ Client TS context improvements

rhyslbw:
This PR addresses a few issues related to connection handling, which was the result of a misunderstood process to establish a connection with the service. To straighten this out I needed to refactor some awkwardness with the connection configuration, which is why we have a large diff. The commits have well defined scope and should be reviewed individually.

# Primary changes
- The client creation promises will now reject if Ogmios is not ready to serve clients, using a new function `getOgmiosHealth` to first check the value of  `lastTipUpdate`. If it's not `null` then a WebSocket is created, (new) onClose and onError handlers attached, before resolving the promise. If it is `null`, then the promise is rejected with a custom error, avoiding the need to use message listeners with what will be a likely immediate closure.
- As mentioned, implementations now must also pass `onClose` and `onError` handlers to gracefully manage disconnects and errors from an established context
- `getOgmiosHealth` is exported and available in the REPL.

```TypeScript
ogmios> await getOgmiosHealth()
{
  lastTipUpdate: '2021-06-19T05:25:38.249991921Z',
  startTime: '2021-06-19T01:55:31.640445805Z',
  metrics: {
    activeConnections: 2,
    totalConnections: 4,
    totalUnrouted: 0,
    sessionDurations: { mean: 48464.5, min: 0, max: 0 },
    runtimeStats: {
      currentHeapSize: 682,
      gcCpuTime: 6648677032,
      cpuTime: 9783746409,
      maxHeapSize: 814
    },
    totalMessages: 110
  },
  currentEra: 'Mary',
  lastKnownTip: {
    hash: 'a4af4646063ecb790d01af4f455078675f2ee294ac05c3c7d31778adb00ea1bc',
    slot: 32514046,
    blockNo: 5870164
  },
  networkSynchronization: 0.99999
}


```

# Todo
- [ ] Add test coverage for the scenario of Ogmios not ready, and the `onClose` and `onError` handlers


clients/TypeScript/packages/client/src/OgmiosHealth.ts
```diff
@@ -0,0 +1,48 @@
...
+  currentEra: 'Alonzo' | 'Byron' | 'Mary' | 'Shelley'
+  lastKnownTip: Tip,
+  lastTipUpdate: string | null,
+  metrics: {
+    runtimeStats: {
+      gcCpuTime: number,
+      cpuTime: number,
+      maxHeapSize: number,
+      currentHeapSize: number
+    },
```
KtorZ, line 20: The runtime stats are only available if run with the right runtime flags (which is the case by default). I don't think there's a particular issue here though, but it could lead to confusing error messages.

rhyslbw: I'll update the type def

rhyslbw: Is the option for deactivating `+RTS -T -RTS`? The [docs](https://ogmios.dev/getting-started/monitoring/) say it's enabled using `+T` but is enabled by default, but the example uses `-T` to "run Ogmios with runtime stats **activated**." Is this a mistake?

I tried to adding it to [cardano-node-ogmios](https://github.com/CardanoSolutions/ogmios/blob/master/scripts/cardano-node-ogmios.sh#L29-L32) :

``` 
ogmios \
  --host 0.0.0.0\
  --node-config /config/cardano-node/config.json \
  --node-socket /ipc/node.socket \
  +RTS -T -RTS
  ```
  ... but am still seeing the `runtimeStats`

Comment (KtorZ) [reply clients/TypeScript/packages/client/src/OgmiosHealth.ts:20]: +RTS -T -RTS is for activating the runtime stats. Passing +RTS -RTS should deactivate them.

Comment (rhyslbw) [reply clients/TypeScript/packages/client/src/OgmiosHealth.ts:20]: I'm still seeing runtime stats with this command:
```
ogmios \
  --host 0.0.0.0\
  --node-config /config/cardano-node/config.json \
  --node-socket /ipc/node.socket \
  +RTS -RTS
```
``` json
{
   "severity":"Info",
   "timestamp":"2021-07-04T02:08:11.003115346Z",
   "thread":"17",
   "message":{
      "tag":"OgmiosHealth",
      "healthCheck":{
         "tag":"HealthTick",
         "status":{
            "startTime":"2021-07-04T02:06:40.667217866Z",
            "lastKnownTip":{
               "headerHash":"306480300948caaa509e4b2dd44a80818ecddc1bccd41513eea9a3af713d4bd2",
               "slotNo":51655,
               "blockNo":50625
            },
            "lastTipUpdate":"2021-07-04T02:08:11.002889343Z",
            "networkSynchronization":0.01683,
            "currentEra":"Byron",
            "metrics":{
               "activeConnections":0,
               "totalConnections":0,
               "totalUnrouted":0,
               "sessionDurations":{
                  "mean":0,
                  "min":0,
                  "max":0
               },
               "runtimeStats":{
                  "currentHeapSize":338,
                  "gcCpuTime":113789651,
                  "cpuTime":4979847237,
                  "maxHeapSize":339
               },
               "totalMessages":0
            }
         }
      }
   },
   "version":"v4.0.0-??-????????"
}
```
I'll just make `runtimeStats` an optional property anyway if that's the intent.

KtorZ:
I am not sure to fully get the rationale for the error handling here. It seems to get quite complicated where I would think it would suffice to wait / retry an only resolve the promise once the service is available?

rhyslbw:
>  I would think it would suffice to wait / retry an only resolve the promise once the service is available?

ie Resolve the promise as soon as the WebSocket service is available? The issue is that until Ogmios registers a tip from the node, the socket would immediately be closed and requires an `onClose` handler for this specific scenario, further complicating the handling for genuine scenarios. To avoid this we're just using the health information to only attempt a connection if this initialisation phase is complete, otherwise reject and allow the implementation to implement retry logic.

clients/TypeScript/packages/client/src/StateQuery/StateQueryClient.ts
```diff
@@ -47,119 +48,118 @@ export interface StateQueryClient {
...
+    socket.once('message', (message: string) => {
+      const response: Ogmios['AcquireResponse'] = JSON.parse(message)
+      if (response.reflection.requestId !== requestId) { return }
+      if ('AcquireSuccess' in response.result) {
+        return resolve({
+          context,
+          currentEpoch: () => {
+            ensureSocketIsOpen(socket)
+            return currentEpoch(context)
+          },
```
KtorZ, line 74: This has nothing to do with this PR per se but just caught my attention :thinking: ...
(and I am pretty sure I raised this in the past and your answer was: doesn't matter because a new client is created for every request)

The state query client return here is "static", so it would only works for a certain time, until the acquired point expired. It would not be a major issue if requests were merely fired immediately, but by returning an intermediate object which does not re-acquire between requests we may end up in bad situations? What mechanisms are given to the client to cope with that? Is it simply that clients are encouraged to create a new client for each request :thinking: ?

clients/TypeScript/packages/client/test/OgmiosHealth.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    it('fetches the service metadata when passed a ConnectionObject', async () => {
+      const connection = await createConnectionObject({ port: 1338 })
+      expectHealth(await getOgmiosHealth(connection))
+    })
+    it('throws fetch errors if encountered', async () => {
+      expect.assertions(1)
+      try {
+        await getOgmiosHealth({ host: 'non-existent' })
+      } catch (error) {
+        expect(error.code).toEqual('EAI_AGAIN')
```
KtorZ, line 31:
```suggestion
        expect(error.code).toMatch(/EAI_AGAIN|ENOTFOUND/)
```

On Linux, the error code is different.

clients/TypeScript/packages/client/src/StateQuery/StateQueryClient.ts
rhyslbw: #87

PR #80, opened 2021-06-29
Chore/docker hub tag rework

rhyslbw:
:pushpin:  **remove superfluous ENV export in Dockerfile**

:pushpin: **tag and push mainnet labelled tags**
Aligns the network tagging strategy to make it easier to dynamically
configure the instance.

`latest`, `latest-{NETWORK}`, `v*.*.*`, `v*.*.*-{NETWORK}`

The existing default tags point to the `mainnet` images, e.g.
`latest` == `latest-mainnet`.

:pushpin: **improve the documented volume name for cardano-node-ogmios**

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         os: [ ubuntu-20.04 ]
         target: [ ogmios, cardano-node-ogmios ]
-        network: [ "", "-testnet" ]
+        network: [ "mainnet", "testnet" ]
 
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yaml
```diff
@@ -39,30 +39,57 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
 
-    - name: 🔨 Build and push
+    - name: 📝 Base Variables
+      id: base-variables
+      run: |
+        echo ::set-output name=image::cardanosolutions/${{ matrix.target }}
```
KtorZ, line 45: There isn't much benefits defining a variable for this one. On [hydra-poc](https://github.com/input-output-hk/hydra-poc/blob/master/.github/workflows/docker.yaml#L13) I ended up configuring the repository in the build matrix itself to use as such:

```yaml
[...]
        cache-from: type=registry,ref=${{ matrix.repository }}:${{ matrix.target }}-latest
[...]
```

But, it doesn't really cost much either to have a variable defined so :man_shrugging:

Dockerfile
```diff
@@ -104,8 +104,6 @@ ENTRYPOINT ["ogmios"]
 FROM debian:buster-slim as cardano-node-ogmios
 
 ARG NETWORK=mainnet
-# Temporary for backwards compatibility.
-ENV NETWORK=$NETWORK
```
KtorZ, line 108: Whoops, my oversight :+1:

KtorZ (approved):

.github/workflows/docker.yaml
rhyslbw: Ah, that's clean. I'll update it

rhyslbw: Ah that's clean, I'll update next time

PR #91, opened 2021-07-27
TypeScript user-guide and client user-facing revisions

KtorZ:
- :round_pushpin: **Add yarn script to re-generate tags file.**
  
- :round_pushpin: **Move socket management logic from 'send' helper into 'InteractionContext'**
    This is a first refactoring step which leaves the outer API unchanged, but will allow to more easily remove the
  use of raw 'ConnectionConfig'. The idea is make the overall API more orthogonal and provide only one way to
  deal pass connection configuration (via the interaction context).

- :round_pushpin: **Make interfaces require 'InteractionContext' everywhere**
    Before this commit, interfaces would accept either a
  'ConnectionConfig' or an 'InteractionContext', but truth that we know
  that the interaction context is safer and likely what one wants to use
  in real system. So, instead of carrying around this double interface
  which makes it more complicated for users to consume the library
  downstream, I've normalized all interface to always work from an
  'InteractionContext'.

  This makes the flow of creating any client much clearer. One pitfall
  however with the current InteractionContext design is that, re-using a
  context across multiple clients may result in odd behavior (e.g.
  calling client.shutDown() may render another client inoperent).

- :round_pushpin: **Remove invalid TODO note**
    proposed protocal parameters are only non-null when protocol updates are submitted, which basically never happen on testnets. Usually, new testnets are re-spinned from scratch instead. So it'll not be feasible, nor practical to observe in tests anyway.

- :round_pushpin: **Export errors related to underlying clients in each client's entrypoint.**
  
- :round_pushpin: **Rework package.json to use multiple entry-points.**
    This allows downstream consumer to import only the bits they're interested in, for example:

    import * as StateQuery from '@cardano-ogmios/client/StateQuery'

- :round_pushpin: **Rename 'release' into 'shutdown' for consistency with other clients.**
  
- :round_pushpin: **Start documenting TypeScript client.**
  
- :round_pushpin: **Continue documenting TypeScript client, with documentation generation via TypeDoc.**
  
- :round_pushpin: **Add @category & @internal annotations to all relevant functions in the client library.**
  
- :round_pushpin: **Rename 'OgmiosHealth' into 'ServerHealth'**
    It sounds clearer like this. In some way, the client is also 'Ogmios'. So here, the role of 'Server' is rather unambiguous and feels more appropriate / self-explanatory

- :round_pushpin: **Adjust REPL interface to also use interaction context instead of mere connection.**
  
- :round_pushpin: **Revert using 'exports' in package.json**
    It's a rather new feature, not well supported it seems. At least jest & typedoc are chocking on it.

- :round_pushpin: **Replace current API reference with generated TypeScript documentation.**
  
- :round_pushpin: **Fix examples in TypeScript user-guide**
  
- :round_pushpin: **Move doc to this repository, packing the theme as a tar.gz archive.**
  
- :round_pushpin: **Adjust getting-started instruction & fix typos in the user-guide**
  
- :round_pushpin: **Define script for installing doc generation pre-requisites.**
  
- :round_pushpin: **Wrap-up user-guide about the TypeScript client.**


clients/TypeScript/packages/client/src/Connection.ts
```diff
@@ -18,8 +18,19 @@ export interface Connection extends Required<ConnectionConfig> {
...
 
+/**
+ * Describe how the interaction context behaves. A `LongRunning` context does not close
+ * the underlying connection after a request, it has to be done manually. A `OneTime` context
+ * however will close the connection afterwards.
+ */
+export type InteractionType = (
+  | 'LongRunning'
+  | 'OneTime'
+)
```
KtorZ, line 32: Now that I've refactored a few things, I also come to doubt the usefulness / relevance of this. Having special logic to handle one-time requests and automatically close connections after sending requests seems a bit superfluous. Rather, this is something that any client's user can do by manually calling `client.shutdown()` when done.

rhyslbw: The driver of this logic is to enable the use of client-less interactions when making state queries or sending transactions.

rhyslbw:
I'll pull this branch and submit a review once I've played with it in an implementation.

clients/TypeScript/packages/client/src/Connection.ts
KtorZ: I see. In that case, I would actually rather define that as a helper function `once` to wrap another query and take care of the resource acquisition for the connection. I'll draft something out.

KtorZ:
Some documentation preview:

Screenshot from 2021-07-29 20-59-45
Screenshot from 2021-07-29 20-59-54
Screenshot from 2021-07-29 21-00-12
Screenshot from 2021-07-29 21-00-30

clients/TypeScript/packages/client/src/Connection.ts
KtorZ: (likely tomorrow or over the weekend :sweat_smile: ....)

rhyslbw (changes requested):
Thanks @KtorZ, excellent work. I agree `InteractionContext` is a fine interface without simplifying it further, and the module restructuring and documentation are great improvements. I'm happy to see this merged once a the minor change requests are addressed

PR #96, opened 2021-08-02
(Slightly Better) Isomorphic WebSocket

KtorZ:
I noticed this while trying to actually use the TypeScript client in a browser context and went digging a bit into this `isomorphic-ws` module. Spoiler: it's a lie! The module is merely a switch which selects the right import at runtime. Yet, it does not attempt to fill the gaps between the Node.js and the Browser-base WebSocket.

The main issue being that, on Node.js, the WebSocket is an instance of [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter) which comes with fairly useful methods like `once`, `removeAllListener` and so forth. On the browser however, we are doomed with the [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) and its crappy API :'( ...

One particularly surprising thing is that, Puppeteer and the supposed cross-platform testing isn't any useful here since Puppeteer seems to be using its own emulation of the WebSocket, which isn't at all the one used by the Browser but has an API closer to the Node.js one. So while the browser tests are all passing, they do not actually pass on a real browser :facepalm:

https://github.com/puppeteer/puppeteer/issues/2119#issuecomment-380961956

This PR introduces a slightly better `IsomorphicWebSocket` interface as a drop-in replacement for our internal use. It only covers the `on`, `once`, `removeListener` and `removeAllListeners` which we use internally.

I had to resort to a JavaScript module for that because I couldn't get the TypeScript compiler to cooperate. As a consequence, the .js module does not get copied in the `dist` by default, I had to manually copy it as part of the build command, which _seems wrong_ but I am too unfamiliar with the TypeScript tooling :/ ... @rhyslbw , happy if you can have a look to avoid this `copy` in the `build` script :/


rhyslbw:
Can you provide the TypeScript source?

KtorZ:
I do not have it around anymore. But it's relatively easy to translate back I guess.

rhyslbw:
Ok, no worries. I can sort it out

PR #110, opened 2021-09-01
Re-acquire state query points

### LINKED ISSUE #90: Provide a way to re-acquire new points in the StateQueryClient

With the Local State Query protocol, queries are run against a specific point that has to be acquired upfront. Yet, acquired points can "expired" and become _too old_. It should be possible to re-acquire new points on demand. 

Similarly, it would also be convenient to have the client automatically re-acquire the most recent point (i.e. the tip) if no point was provided initially. Note that, this is also the behavior of the server at the moment which does not require client to explictely _acquire_. Clients are free to send queries immediately, in which case they are executed against the latest tip. So all-in-all, implementing this could be as simple as avoiding the initial `Acquire` when creating the state query client. 

#### Why is it a good idea?

Right now, the only work-around for an app would be to re-create a new client with a more recent point. That is quite unpractical. 

Maybe.

KtorZ:
- :round_pushpin: **Change state query client setup to not acquire when no point is provided.**
    Behind the scene, the server will automatically acquire the most recent tip and run the query for that point. This however exposes clients to race conditions should they make multiple queries one after the other and expect them to be consistent.

- :round_pushpin: **Allow state query client to re-acquire points when necessary.**

Fixes #90

### DIFF
--- a/clients/TypeScript/packages/client/src/StateQuery/StateQueryClient.ts
+++ b/clients/TypeScript/packages/client/src/StateQuery/StateQueryClient.ts
@@ -25,7 +25,7 @@ import {
   stakeDistribution,
   utxo
 } from './queries'
-import { createPointFromCurrentTip, ensureSocketIsOpen, safeJSON } from '../util'
+import { ensureSocketIsOpen, safeJSON } from '../util'
 
 /**
  * A State Query client.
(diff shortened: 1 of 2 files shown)


rhyslbw (approved):
Nice work!

Issue #89, opened 2021-07-22
client/TypeScript: Fault response ignored


rhyslbw:
#### What Git revision are you using? 
`a7bab073220caad8b98351515e3d68f5eb605383`
#### What operating system are you using, and which version?

- [x] Linux / Ubuntu
- [ ] Linux / Other
- [ ] OSX
- [ ] Windows

#### Describe what the problem is? 
As reported in https://github.com/input-output-hk/cardano-graphql/issues/545
```
> await submitTx('abc')
# hangs
```

#### What should be the expected behavior?
A failure response

rhyslbw:
There are two issues:
1. The server is not reflecting the request mirror value in a `Fault` response, which is used by the client to match handlers to requests.
_Line 30 discards the response_
https://github.com/CardanoSolutions/ogmios/blob/320c97aef8a823e6295412b16a8a3b29bea15dfe/clients/TypeScript/packages/client/src/StateQuery/Query.ts#L28-L37

2. The `Fault` response will fall-through to the 'UnexpectedResponseError`, so that should be improved.
https://github.com/CardanoSolutions/ogmios/blob/320c97aef8a823e6295412b16a8a3b29bea15dfe/clients/TypeScript/packages/client/src/TxSubmission/submitTx.ts#L156

KtorZ:
Note: as per b94a8b2daba8fac55534083a3de8a96337d11339, the server now correctly returns the matching `mirror` / `reflection` on faults for invalid requests.

KtorZ:
Fixed in #116

PR #116, opened 2021-09-08
Handle client faults in the TypeScript StateQueryClient

xdzurman:
Motivation: When you enter an invalid stake address to delegationsAndRewards API endpoint, the function fails on

(node:267433) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'eraMismatch' of undefined
at isEraMismatch (node_modules/@cardano-ogmios/client/src/StateQuery/queries/delegationsAndRewards.ts:16:9)
    at Object.handler (node_modules/@cardano-ogmios/client/src/StateQuery/queries/delegationsAndRewards.ts:46:8)

and never finishes. This fix provides at least a rejection instead of neverending await.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ chapter: false
 pre: "<b>6. </b>"
 ---
 
+### [4.0.1] - 2021-09-08
+
+#### Added
+
+N/A
+
+#### Changed
+
+- The TypeScript `StateQueryClient` now wraps every query in a try/catch to cope with malformed queries leading to client `fault` results from the server.
+
+#### Removed
+
+N/A
+
+
 ### [4.0.0] - 2021-09-06
 
 #### Added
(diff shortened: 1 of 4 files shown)


KtorZ:
That's a good point. The server actually returns a 

```json
{
  "fault": {
    "string": "Invalid request: invalid query.",
    "code": "client"
 }
```

Which isn't handled properly in the TypeScript client. And I'd imagine likely the case for other queries as well so I'd go for a more "holistic" approach and introduce the error handler in `StateQuery/Query.ts` directly: 

```ts
        async function listener (data: string) {
          const queryResponse = safeJSON.parse(data) as QueryResponse
          if (queryResponse.reflection?.requestId !== requestId) { return }
          try {
            await response.handler(
              queryResponse,
              resolve,
              reject
            )
          } catch (e) {
            return reject(new UnknownResultError(queryResponse))
          }
          socket.removeListener('message', listener)
        }
```

This is related to (and would fix) #89

KtorZ (approved):

Issue #120, opened 2021-09-15
Possible memory leak in ogmios client implementation


MarcelKlammer:
#### What Git revision are you using? 

Using docker image cardanosolutions/ogmios:-v4.0.0 for ogmios service.

#### What operating system are you using, and which version?

Linux / Ubuntu: 20.04 LTS

#### Describe what the problem is? 

We implemented https://github.com/CardanoSolutions/ogmios/tree/master/clients/TypeScript/packages
client and schema packages to call StateQueryClient.delegationsAndRewards for stake address balances.

After deploying the client to our backend and getting some load (200+ requests per minute) we saw a steady increase in memory consumption. We made sure, that is indeed ogmios that's causing the increase.

We started with having 1 client per network (guild, testnet, mainnet), not shutting it down, because with 200 calls it seemed wastful. The memory increase was large. We then opened a client per request and called shutdown each time. The memory increase was slower than before, but also too big to let it slide on our Heroku instances.

#### What should be the expected behavior?

Memory should not increase over time under moderate load.

KtorZ:
@MarcelKlammer I've been unable to reproduce so far. I've tried some simple stress test with a `setInterval` loop and several scenarios:

- Creating a single `LongRunning` client outside of the loop, and using it to make many calls
- Creating a new `LongRunning` client inside the loop, and purposely not shutting it down.
- Creating a new `OneTime` client inside the loop.

I left it running for a few minutes without observing any issue on memory usage (both the server and the node process). If you have any pointers on how to reproduce this, that'd be helpful.

KtorZ:
@MarcelKlammer did you find anything about this in the end? Or may I close the issue :thinking:

MarcelKlammer:
Running newer versions for a while, nothing to report. Closed.

PR #123, opened 2021-09-18
fix: address floating promises

rhyslbw:
In an attempt to narrow #120 I discovered we have a few floating promises. This doesn't appear to touch the code of interest, so is unlikely to change much there, but it should improve the reliability of tests given the `delay`s were not being awaited.


KtorZ (approved):
👌

PR #126, opened 2021-09-30
Test: client typescript test file restructure

rhyslbw:
Due to the implementation of TypeScript path mapping, all test files have to be directly under `test`, which looses structure determined by the source. I also found Puppeteer was being run in non-headless mode.

:pushpin: **test: restructure test files to mirror the source**
This comes at the cost of removing path mapping, which is incompatible
with the current tooling.

:pushpin: **test: run Puppeteer in headless mode**

image


clients/TypeScript/packages/client/test/ChainSync/ChainSync.test.ts
```diff
@@ -1,7 +1,7 @@
 import {
   ChainSyncMessageHandlers,
   createChainSyncClient
-} from '@src/ChainSync'
+} from '../../src/ChainSync'
```
rhyslbw, line 4: Without aliased path mapping

KtorZ (approved):
Nice!

PR #134, opened 2021-10-25
Feat/client TS abstract logger

### LINKED ISSUE #69: client/TypeScript: Abstract Logging
I have a bunch of generic logging at the application level of my current project that best belongs upstream. I would like to hoist the abstract calls and refactor the clients to accept a logger instance compatible with `console` or [bunyan](https://www.npmjs.com/package/bunyan).

- The abstract logger [ts-log](https://www.npmjs.com/package/ts-log)  would be the added dependency
- Default operation is silent.

#### Why is it a good idea?
To increase the value of the clients with consistent logging behaviour across implementations. It's valuable for both development and production as an opt-in feature.

Yes

rhyslbw:
Closes #69

:pushpin: **test: disable impractical unbounded utxo test**
This is a valid query based on the Ouroboros mini-protocol, however as it lacks batching, the unbounded invocation is impractical for larger networks. Disabling here since we're using the `testnet`.

:pushpin: **fix: getServerHealth connection param**
Connection param for `getServerHealth` is now required, reflecting
the function's dependency on the value being passed in. While the
interface is changing, it's actually better considered a fix since
omitting the argument would throw an error.

:pushpin: **feat: client TypeScript abstract logging**
Introduces an abstract logger compatible with `console` or more sophisticated
loggers such as bunyan, with the default operation being silent. Relevant
interfaces now accept an optional logger argument.

_Draft until I've done an integration to get a feel for logging points given this is a subjective issue._

### DIFF
--- a/clients/TypeScript/packages/client/package.json
+++ b/clients/TypeScript/packages/client/package.json
@@ -59,6 +59,7 @@
     "isomorphic-ws": "^4.0.1",
     "nanoid": "^3.1.22",
     "ts-custom-error": "^3.2.0",
+    "ts-log": "^2.2.3",
     "ws": "^7.4.6"
   },
   "files": [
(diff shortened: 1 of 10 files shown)


KtorZ:
Closing for now, feel free to come back with this later. I've however included the option breaking change for the health function in 5.0.0 such that hopefully, introducing the abstract logger won't be a breaking change!

PR #138, opened 2021-11-06
Fix a broken link in Docker.md

bjing:


KtorZ:
Hey, good catch! I've also caught another one while reviewing your PR, and actually pointed both new links to the _actual source of truth_ which is the Github workflows.

KtorZ (approved):

PR #162, opened 2022-01-16
Local Tx Monitoring

KtorZ:
- :round_pushpin: **Upgrade dependencies to cardano^1.33.0 + support for local transaction monitoring.**


server/src/Ogmios/Data/Json/Orphans.hs
```diff
@@ -62,6 +64,9 @@ instance ToJSON (Point (CardanoBlock crypto)) where
 instance PraosCrypto crypto => FromJSON (GenTx (CardanoBlock crypto)) where
     parseJSON = decodeSubmitTxPayload
 
+instance FromJSON (GenTxId (CardanoBlock crypto)) where
+    parseJSON = error "FromJSON: GenTxId"
```
KtorZ, line 68: TODO :point_up:

Issue #194, opened 2022-03-21
DOMException: An invalid or illegal string was specified


siegfried:
I cannot create a `InteractionContext` in browser (Latest Firefox and Chrome). After a while of troubleshooting, I believe the issue lies [here](https://github.com/CardanoSolutions/ogmios/blob/7ada4945fe205c94a5406dd9e39848bf428a000c/clients/TypeScript/packages/client/src/Connection.ts#L92).

I tried this successfully in browser console:

```javascript
new WebSocket('ws://localhost:1338')
```

This gives the error: `Uncaught DOMException: An invalid or illegal string was specified`

```javascript
new WebSocket('ws://localhost:1338', { maxPayload: 128 * 1024 * 1024 })
```

KtorZ:
Eew. Yes, the so-called "isomorphic" Javascript WebSockets aren't quite isomorphic. We resorted to building [our own wrapper around the websocket API](https://github.com/CardanoSolutions/ogmios/blob/master/clients/TypeScript/packages/client/src/IsomorphicWebSocket.ts) to cover both browser and node. Note however that the browser support in the TypeScript client is only a best effort as Ogmios is mostly intended to be used server-side in a controlled environment.

siegfried:
So is Ogmios not supposed to be exposed to the public?

siegfried:
I took a deep look if I can provide some fixes. Where is [isBrowser](https://github.com/CardanoSolutions/ogmios/blob/7ada4945fe205c94a5406dd9e39848bf428a000c/clients/TypeScript/packages/client/src/IsomorphicWebSocket.ts#L9) from BTW? Is it ok to use a package like this? https://www.npmjs.com/package/browser-or-node

KtorZ:
@siegfried in principle _no_; Ogmios exposes the cardano-node's node-to-client protocols which have been designed to be **local** protocols between a limited set of clients and one node acting as a server. Some protocols are quite heavy on the node (for example, the chain-sync protocol can drive the node at 200% CPU for an extended period of time). 

This isn't a security risk though, but more of a resource constraint issue. We've been experimenting having hundreds of clients connected to a node. And while this _works_, it is quite slow (especially if all clients are doing chain-sync). So in general, no, exposing an instance over the public internet is not a good idea unless you are in a somewhat controlled environment or, unless there's way to spread the load over many relays. 

[...]

PR #195, opened 2022-03-22
Skip options for browser WebSocket()

siegfried:
Could just skip the second `protocols` argument of `WebSocket` constructor, which accepts a `string` or `string[]`. Here are [details of the value](https://javascript.info/websocket#extensions-and-subprotocols).


siegfried:
This is related to #194

KtorZ:
Thanks for looking into this :+1: , it is quite unfortunate that the pupeteer toolsuite isn't able to catch that... I had a look a while back but even when running in a browser context, it seems like [pupeteer offers a non-compatible websocket API](https://github.com/puppeteer/puppeteer/issues/2119#issuecomment-380961956) and resort to the node.js interface. I haven't checked whether recent versions fixed the issue or not. 

So anyway, thanks :+1:

KtorZ (approved):

PR #214, opened 2022-05-16
Local tx monitor typescrypt client

leobel:
Add typescript client for LocalTxMonitor Mini-Protocol


clients/TypeScript/packages/client/src/LocalTxMonitor/LocalTxMonitorClient.ts
```diff
@@ -0,0 +1,195 @@
...
+        AsyncGenerator<Ogmios['HasTxResponse']>
+
+    const nextTxResponse = eventEmitterToGenerator(socket, 'message', matchNextTx)() as
+        AsyncGenerator<Ogmios['NextTxResponse']>
+
+    const sizeAndCapacityResponse = eventEmitterToGenerator(socket, 'message', matchSizeAndCapacity)() as
+        AsyncGenerator<Ogmios['SizeAndCapacityResponse']>
+
+    const releaseMempoolResponse = eventEmitterToGenerator(socket, 'message', matchReleaseMempool)() as
+        AsyncGenerator<Ogmios['ReleaseMempoolResponse']>
```
KtorZ, line 94: That's one way of doing it :sweat_smile: ... Intuitively, I would have gone for a single generator matching all the messages. This could reduce a bit the boilerplate here I reckon and would be fine because the server handles messages sequentially behind the scene :thinking: 

I know I did use two event queues in the TxSubmission protocol but in principle, there was no need and I guess I was didn't do this thought experiments.

clients/TypeScript/packages/client/src/LocalTxMonitor/awaitAcquire.ts
```diff
@@ -0,0 +1,47 @@
...
+        const { result } = response
+        if ('AwaitAcquired' in result) {
+            return result.AwaitAcquired
+        } else {
+            return [new UnknownResultError(response)]
+        }
+    } catch (e) {
+        return [new UnknownResultError(response)]
+    }
+}
```
KtorZ, line 47: There is really no error expected as result of this message (and actually, to any messages of this protocol); so we could throw earlier to simplify the response handling and manipulation of the response quite a bunch:

```ts
export const handleAwaitAcquireResponse = (response: Ogmios['AwaitAcquireResponse']): AwaitAcquired => {
    const { result } = response
    if ('AwaitAcquired' in result) {
        return result.AwaitAcquired
    } 
    throw new UnknownResultError(response)
}
```

Which leads to:

```ts
handler: (response, resolve, reject) => 
    try {
        resolve(handleAwaitAcquireResponse(response))
    } catch (e) {
        reject(e)
    }
```

And similarly, in the client definition:

instead of 

```ts
const response = handleAwaitAcquireResponse((await awaitAcquireResponse.next()).value)

if (isAwaitAcquiredResult(response)) {
    return response
} else {
    throw response
}
```

we can write

```ts
return handleAwaitAcquireResponse((await awaitAcquireResponse.next()).value)
```

clients/TypeScript/packages/schema/src/index.ts
```diff
@@ -8,53 +8,11 @@
 /**
  * A Cardano Block, which can take different forms depending on the era.
  */
-export type Block = Byron | Shelley | Allegra | Mary | Alonzo;
```
KtorZ, line 11: I am not sure to understand why there are so many changes here :thinking: ? Wasn't the schema up-to-date already? Or perhaps did you use a more recent version of json-schema-to-typescript?

Any clue?

KtorZ (commented):
Firstly, thanks a lot for taking the time going through this! This was on my TODO list for a while but so far, no one had really requested it so :shrug: 

I left a few comments, mostly renaming but also things that I got to realize just now and could also be improved in the TxSubmission client. 

I haven't read through the test just yet; I'll ASAP.

PR #220, opened 2022-05-28
Vasil/Babbage: 1.35.0-rc1 Upgrade

### LINKED ISSUE #215: Vasil hardfork compatibility
Apologies if I've missed something, but I wanted to ask if there is an existing plan to upgrade Ogmios for compatibility with features introduced after the Vasil hardfork. I couldn't find any issues or milestones referencing this

We use Ogmios extensively and will start Vasil testnet testing fairly soon. We would be happy to try to upstream any changes we need to make to Ogmios if there's no existing upgrade plan

Thank you for creating Ogmios!

KtorZ:
Fixes #215

---

- :round_pushpin: **Upgrade dependencies to cardano^1.35.0-rc1**
  
- :round_pushpin: **Upgrade json-wsp to use aeson >= 2.0.0.0.**
  
- :round_pushpin: **Add babbage client codecs.**
  
- :round_pushpin: **Start upgrading Ogmios to Vasil/Babbage.**
    The code now 'compiles' without -fproduction; There are still many incomplete pattern matches and 'undefined' but, they should only be relatively straightforward updates. The heavy lifting was done in the 'TxSubmission' protocol and in particular the execution units evaluation which needed to be made era-independent. I've also reworked a bit the deserialization code for transaction so that it is less awkward to read and actually provides some degree of type-safety.

   Another large chunk remaining will be the 'Ogmios.Data.Json.Babbage' module to cope with newly introduced elements in babbage.

- :round_pushpin: **Upgrade queries, JSON parsers and remaining glue code.**
    Now only truly remains Babbage's JSON support.

  ... and testing. I haven't tried to even compile the test-suite yet.
  I expect some unhappy breakage here.

- :round_pushpin: **Support Babbage types in JSON decoders**
    The decoder for transaction outputs is now a bit more flexible to cope
  with breaking-changes on previous versions. Transaction outputs can
  now have either a `datum` or a `datumHash`. They can also have a
  `script`.

  For parsing `script`, I've made a quite conservative choice of
  allowing two distinct formats (this needs to be documented!). It
  accepts either scripts as serialized on-chain in raw transactions, or
  a JSON-wrapped version compatible with what the server returns. I am
  not sure _yet_ whether the first format is needed and need to play
  around a bit with existing tools and double-check what both the
  plutus toolkit and the cardano-api typically spit back.

- :round_pushpin: **Begin definitions of JSON types for Babbage.**

- :round_pushpin: **Wrap up remaining JSON encoders for Babbage**
    Noticed something a bit weird in the Babbage predicate failures, so
  opened that:

  https://github.com/input-output-hk/cardano-ledger/issues/2843

- :round_pushpin: **Update hjsonpointer to support aeson > 2.0.0.0**
  
- :round_pushpin: **Compile test-suite and fix warts JSON-schema specifications.**
  
- :round_pushpin: **Add NodeToClientV_13 to the list of supported network protocols.**
  
- :round_pushpin: **Remove hie.yaml, should never have been committed to begin with.**
  
- :round_pushpin: **Bump cardano-node docker versions to 1.35.0-rc1**

- :round_pushpin: **Drop unused language pragma.**
  
- :round_pushpin: **Fix remaining issues with the cabal build + prepare Nix build**
    Hitting a wall here. Seems like cross-compilation with musl64
  doesn't work anymore because of the introduction of libsecp256k1
  (for Schnorr-compatible signature verification builtin in Plutus).

  I've raised this to the IOG's tools team on haskell.nix under:

  https://github.com/input-output-hk/haskell.nix/issues/1505

  :finger_crossed:

- :round_pushpin: **Use nixpkgs-unstable from haskell.nix to resolve issues with secp256k1.**
  
- :round_pushpin: **Build secp256k1 also for the test job in CI workflow.**
  
- :round_pushpin: **Extend the JSON-schema specifications with Babbage types.**
    Mainly, block, transaction and protocol parameters as well as new
  transaction submission and evaluation failures (coming from new ledger
  rules).

- :round_pushpin: **Integrate Vasil/Babbage changes into the TypeScript client.**
    Mostly re-generating the TypeScript types from the JSON-schema, and then following the TS compiler.

- :round_pushpin: **Rename 'collateralReturnMismatch' into 'totalCollateralMismatch'.**
  
- :round_pushpin: **Fill-in CHANGELOG for v5.5.0-rc1**

- :round_pushpin: **Add roundtrip property for Babbage utxo encoding/decoding**
    Much needed. I got it wrong about 8 times I think before actually succeeding. There was many things going wrong, and the dual support for Alonzo/Babbage is still very much annoying.
  Yet, property-test being the most amazing thing given to mankind, I eventually managed to (hopefully) get to a working, robust, version.

- :round_pushpin: **Also install libsecp256k1 in TypeScript CI job**
    The server is actually built in the previous step as a dynamic
  executable. So we need to have the same dependencies installed
  on both VMs.

- :round_pushpin: **Fix packaging step in workflow**
    The nix-build command used to automatically generate a 'share' folder
  hierarchy with the software LICENSE. Seems like more recent versions
  of Haskell.nix does not do that anymore, though the 'share' folder is
  still needed to ship with auto-completions scripts.

- :round_pushpin: **Bump version(s) to v5.5.0(-rc1)**


server/src/Ogmios/Data/Json/Alonzo.hs
```diff
@@ -166,14 +171,20 @@ encodeCollectError = \case
...
+encodeCostModels =
+    encodeMap stringifyLanguage encodeCostModel . Al.unCostModels
 
 encodeData
     :: Al.Data era
     -> Json
-encodeData (Al.DataConstr datum) =
-    encodeShortByteString encodeByteStringBase64 (memobytes datum)
+encodeData (Al.Data datum) =
+    encodeByteStringBase16 (toStrict (serialise datum))
```
KtorZ, line 187: Small breaking change; probably for the better but might revise later.

server/src/Ogmios/App/Protocol/TxSubmission.hs
```diff
@@ -156,26 +186,31 @@ newExecutionUnitsEvaluator = do
...
+                (_, GenTxMary{}) ->
                     return (incompatibleEra "Mary")
-                GenTxAlonzo (ShelleyTx _id tx) -> do
-                    atomically $ putTMVar evaluateExecutionUnitsRequest (utxo, tx)
+                (SomeUTxOInAlonzoEra utxo, GenTxAlonzo (ShelleyTx _id tx)) -> do
+                    atomically $ putTMVar evaluateExecutionUnitsRequest $ SomeEvaluationInAnyEra utxo tx
+                    atomically $ takeTMVar evaluateExecutionUnitsResponse
+                (SomeUTxOInBabbageEra utxo, GenTxBabbage (ShelleyTx _id tx)) -> do
+                    atomically $ putTMVar evaluateExecutionUnitsRequest $ SomeEvaluationInAnyEra utxo tx
                     atomically $ takeTMVar evaluateExecutionUnitsResponse
```
KtorZ, line 203: TODO: could/should translate Alonzo to Babbage here in case of discrepancy. Especially because transactions and utxos will parse as old eras first, so it is possible to submit a Babbage transaction with an Alonzo UTxO. In which case, we'll consider the most recent era -> Babbage.

server/src/Ogmios/App/Protocol/TxSubmission.hs
```diff
@@ -184,102 +219,207 @@ newExecutionUnitsEvaluator = do
...
+          in
+            sameEra <|> alonzoToBabbage
+
+selectEra
+    :: forall block c m.
+        ( block ~ HardForkBlock (CardanoEras c)
+        , Applicative m
+        )
+
+    -- | Default selector, when network era is prior to Alonzo and don't support Phase-2 scripts.
```
KtorZ, line 400: double-typo.

PR #224, opened 2022-06-10
chore: remove unused cardano-configurations submodule

rhyslbw:
This was previously used to load configuration into the `cardano-node-ogmios`
Docker image, but the config is now being [cloned as part of the build definition.](https://github.com/CardanoSolutions/ogmios/blob/8b08ce856c5dda12f7918cd21718a8da5fc4904f/Dockerfile#L15)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "modules/hjsonschema"]
 	path = server/modules/hjsonschema
 	url = https://github.com/KtorZ/hjsonschema.git
-[submodule "server/config"]
-	path = server/config
-	url = https://github.com/input-output-hk/cardano-configurations.git
 [submodule "clients/Go"]
 	path = clients/Go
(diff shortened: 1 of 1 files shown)


KtorZ (approved):

KtorZ:
Approving but, thinking about it, I still also like that the repository contains some configuration files nearby :thinking: . It's quite handy when playing around, in particular outside of docker.

rhyslbw:
Looks like it's used in the server tests anyway, so I can just close this PR. What was the thinking by cloning the latest rather than using the tested local source that's in revision control?

KtorZ:
> What was the thinking by cloning the latest rather than using the tested local source that's in revision control?

I think the issue was more on the side of Nix and, getting the nix-build to work with git submodules :disappointed: ...  I have bad memories from this, I probably gave up at some point and went for just cloning the repository.

PR #234, opened 2022-07-05
Update protocol parameter helpers in TypeScript client

KubqoA:
The helpers to determine what era the protocol parameters belong to were "out-dated".
1. Existing ones (`isShelleyProtocolParameters` and `isAlonzoProtocolParameters`) were hard-to-use, as they didn't accept the new `ProtocolParametersBabbage` type
2. And the function `isBabbageProtocolParameters` was missing.

This PR should fix both of these issues.


KtorZ (approved):
Nice catch. Thanks :pray:

KubqoA:
Thanks for the quick approve, and thanks for Ogmios 🙂

PR #246, opened 2022-08-15
Chore/bump cardano node to 1.35.3 and cardano-configurations to latest

rhyslbw:


KtorZ:
Hmmm. Looks like the Vasil devnet was restarted in August. Needs to fix the options spec.

rhyslbw:
I've added the new testnets in 5f343ee4a34467c091cb4ba04777019f62da7703. For information relating to these, see https://book.world.dev.cardano.org/environments.html

rhyslbw:
We'll also need to drop _testnet_ and ensure TS tests are all passing on _preprod_. I'm out of time tonight, but can sort it tomorrow during my daylight hours.

KtorZ:
> since cardano-node@1.35.3 is no longer able to sync to the tip 

:|

rhyslbw:
I've found a bug in the TS client that needs addressing. `EvaluationFailure[ScriptFailures]` is an [object](https://github.com/CardanoSolutions/ogmios/blob/e2a6ff63e06be3b16065c82a9c0e3be7d51dfca1/docs/static/ogmios.wsp.json#L2732), but it's [expected to be an array](https://github.com/CardanoSolutions/ogmios/blob/e2a6ff63e06be3b16065c82a9c0e3be7d51dfca1/clients/TypeScript/packages/client/src/TxSubmission/evaluateTx.ts#L63), with no `else` clause to handle the exception. Curiously there's been no changes from the original implementation, so it's surprising to only be caught now :thinking:

KtorZ:
Interesting. You only hit evaluation failures though when you explicitly evaluate execution units, not when submitting. So if you've not been using that feature before, you'd have not noticed.

Issue #250, opened 2022-08-27
Build and upload multi-platform Docker images


jehrhardt:
#### Describe your idea, in simple words.
Publish multi-platform Docker images to Docker Hub. Support at least:

- linux/amd64
- linux/arm64

#### Why is it a good idea?
AWS provides Graviton as a cheaper option. Apple provides its laptops with M1 / M2 chips. Multi-platform images can provide support for multiple CPU architectures in a single image.

#### Are you willing to work on it yourself?
🤔 probably yes.

jehrhardt:
The pull requests added arm64 support to the ogmios image, but not to the one bundled with Cardano node. Is this enough? Or should we investigate to build the other for arm64 as well?

KtorZ:
I wonder if it's possible. I haven't tried to build it and I am far away from my laptop for the next week so I can't try. For the sake of consistency and completeness, I'd rather have it if possible. The cardano-node-ogmios' image however is based on the cardano-node upstream image, which may or may not play well with this cross-platform docker builds.

jehrhardt:
Ah, damn. Cardano node upstream image is amdzr only. Maybe we should close this one and follow up, once the upstream image works.

KtorZ:
The build is failing for arm for the base image as well 😔. If you want to have a look, otherwise I'll next week.

wolf31o2:
I publish cardano-node images built from upstream source, but on both AMD64 and ARM64, with the same interface as the upstream IOG images. I was just about to create my own wrapper for this project so I could get working ARM64 images. Would you accept a pull request that switched the images? It would be a drop-in replacement in this usage.

Images:
https://hub.docker.com/r/cloudstruct/cardano-node
Sources:
https://github.com/cloudstruct/docker-cardano-node

PR #251, opened 2022-08-27
Add arm64 as platform to ogmios image build

jehrhardt:
This will trigger a build for both amd64 and arm64. The result is a multi-platform image that works on for both CPU architectures.

refs #250


KtorZ:
Thanks for the PR ❤️
I am just wondering how dockerhub distinguish images and whether we need to tag them separately?

KtorZ:
It seems like this is handled directly by the server and the pull command. It's during a docker pull that one specifies which platform it needs.

KtorZ (approved):

PR #252, opened 2022-08-27
Fix the format of multi-platform build config

jehrhardt:
Looks like miss-interpreted the docs. The platforms option needs to be provided differently.

refs #250 #251


KtorZ (approved):
Arf,  indeed.

PR #258, opened 2022-09-20
Fix mix typo in metrics

sorki:
.. min and max got mixed


KtorZ (approved):
Arf. Indeed, somehow didn't catch that during review..

PR #301, opened 2023-02-01
Fix 'sanitize' to support Value without 'assets'

mkazlauskas:
`Value` without assets comes as `{coins:n}` instead of `{coins:n, assets:{}}` from the server, which results in `Lovelace` not being sanitized.

### DIFF
--- a/clients/TypeScript/packages/client/src/util.ts
+++ b/clients/TypeScript/packages/client/src/util.ts
@@ -33,12 +33,16 @@ export const safeJSON = {
     if (typeof json === 'object' && json !== null) {
       const len = Object.getOwnPropertyNames(json).length
 
-      // AssetQuantity
-      if (len === 2 && json.coins !== undefined && json.assets !== undefined) {
+      // Lovelace & AssetQuantity
+      if (json.coins !== undefined) {
         const coins = json.coins
         json.coins = typeof coins === 'number' ? BigInt(coins) : coins
 
-        return this.sanitizeAdditionalFields(json.assets)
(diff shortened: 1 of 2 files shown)


KtorZ:
Hey @mkazlauskas, good catch.

> Value without assets comes as {coins:n} instead of {coins:n, assets:{}} from the server

They actually do not -- except for values coming before the Mary ere because they had no `assets` whatsoever... As discussed with @rhyslbw, the current deserialization work-around is not ideal because there's no good ways to do any kind of type-reflection at runtime. A more robust alternative would be to deserialize JSON data alongside a JSON-schema which would contains type information to use. I never bothered doing that in [CardanoSolutions/json-bigint](https://github.com/CardanoSolutions/json-bigint) yet I imagine this to be relatively quick to implement.

clients/TypeScript/packages/client/src/util.ts
```diff
@@ -33,12 +33,16 @@ export const safeJSON = {
...
-      // AssetQuantity
-      if (len === 2 && json.coins !== undefined && json.assets !== undefined) {
+      // Lovelace & AssetQuantity
+      if (json.coins !== undefined) {
         const coins = json.coins
         json.coins = typeof coins === 'number' ? BigInt(coins) : coins
 
-        return this.sanitizeAdditionalFields(json.assets)
+        if (json.assets !== undefined) {
+          return this.sanitizeAdditionalFields(json.assets)
```
KtorZ, line 42: Good catch too.

KtorZ (approved):

PR #345, opened 2023-10-21
Add SanchoNet variant to Dockerhub builds

rhyslbw:
SanchoNet is a public network that will be supported for some time. It would be great to have pre-built images while this network is live


KtorZ:
Damn. I thought I already did that. Good catch, thanks.

PR #353, opened 2023-11-23
Update sha256 for ouroboros-network

angerman:
The one listed in the project appears to be incorrect and not match the revision.


aciceri:
@angerman This PR let me think that you build ogmios with `haskell.nix`. Is this deduction correct? In a public repo?

PR #362, opened 2024-01-25
Fix extracting method from response

mirelon:
When sending two requests to ogmios, e.g. `protocolParameters` and `evaluateTransaction`, response for `protocolParameters` is matched in `matchEvaluateTransaction` because `'method' in json.id` is false.

There is a top-level `method` field as defined in https://ogmios.dev/api/#operation-publish-/?EvaluateTransaction


KtorZ:
Ho, Nice catch. This looks about right although I need to double check how that passed through the test. The id is how the pre-v6 client used to work and I don't recall changing anything there, so it should still be actually working.

Issue #363, opened 2024-01-31
evaluateTransaction dropping connection on sanchonet


klarkc:
### What Git revision / release tag are you using?

v6.0.2

### Do you use any client SDK? If yes, which one?

cardano-transaction-lib

### Describe what the problem is?

I'm attempting to evaluate this transaction:

```json
{"params":{"transaction":{"cbor":"84a4008182582055ed5376ddaafcc1d498d96d7d45ebda05a4c63c3f9e0e99a345df0218db8cb103018283583910793f8c8cffba081b2a56462fc219cc8fe652d6a338b62c7b134876e7899caf87c60e581582a408206fe6cc58e7a779afbc516b3726a82f2e1a001e84805820923918e403bf43c34b4ef6b48eb2ee04babed17320d8d1b9ff9ad086e86f44ec82583900a838a4e1280f3bf8e1462f7a9d728b3e277427cd08d0512594603ccd899caf87c60e581582a408206fe6cc58e7a779afbc516b3726a82f2e1a4a4fe53002000f00a2049fd87980ff0580f5f6"},"additionalUtxo":[]},"method":"evaluateTransaction","jsonrpc":"2.0","id":"evaluateTransaction-ls2ebvv3"}
```

The connection is dropping without any additional information on the Ogmios side. Cardano-transaction-lib handles the connection drop by retrying a new connection with payload resubmission. Thus, it becomes an endless loop.

### What should be the expected behavior?

I'm expecting any kind of error (if the CBOR is being incorrectly serialised I was expecting any error as response)

### If applicable, what are the logs from the server around the occurence of the problem?

https://gist.github.com/klarkc/046a6d2c731b979d556f338d1b479ffe

KtorZ:
It's odd. Seems like it is the node closing the connection unexpectedly. This should in principle lead to an error response being sent from Ogmios' side; maybe it is being ignored by `cardano-transaction-lib`? 

Note that I am not able to reproduce this and I am simply getting a response:

```json
{"jsonrpc":"2.0","method":"evaluateTransaction","result":[],"id":"evaluateTransaction-ls2ebvv3"}
```

which seems to make sense since the transaction doesn't carry any scripts for execution? 

```json
❯ ogmios inspect transaction 84a[...]5f6 | jq
{
  "id": "a7da8e2c27fda72ec20d51ac10ea7265b22d6eb67c6b1d14defa5b2aaac95165",
  "spends": "inputs",
  "inputs": [
    {
[...]

PR #383, opened 2024-04-12
Add upgrade unsafeFromRight reporting workaround

klarkc:
I'm encountering an upgrade (https://github.com/CardanoSolutions/ogmios/issues/363#issuecomment-1953067395) decode error again (but for different reasons), and just like before, it's not being reported. Therefore, I'm proposing this change until we establish a better method for reporting decode errors in the upgrade function.

Related #363

### DIFF
--- a/server/src/Ogmios/Data/EraTranslation.hs
+++ b/server/src/Ogmios/Data/EraTranslation.hs
@@ -19,6 +19,7 @@ module Ogmios.Data.EraTranslation
     ) where
 
 import Ogmios.Prelude
+import System.IO.Unsafe (unsafePerformIO)
 
 import Cardano.Crypto.DSIGN
     ( DSIGNAlgorithm (..)
@@ -64,7 +65,8 @@ type family BlockEra block :: Type where
 
 class Upgrade (f :: Type -> Type) era where
     type Upgraded f :: Type -> Type
-    upgrade :: f (PreviousEra era) -> Upgraded f era
+    -- FIXME HasCallStack is not printing the stack on errors (see unsafeFromRight)
+    upgrade :: HasCallStack => f (PreviousEra era) -> Upgraded f era
 
 ----------
 -- TxOut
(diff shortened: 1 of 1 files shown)


KtorZ:
@klarkc not a big fan of the change 😬 ... this kind of temporary fix ends up being more permanent than we usually want initially. 

So I am more for biting the bullet and getting those async exception handled properly. It might just be a matter of linking the async thread with the parent as this should crash the entire connection. 

Do you have an easy reproduction step here by the way? I can always just shove an artificial `undefined` here and there, but I am genuinely curious about what's causing this now?

klarkc:
@KtorZ It happened with ogmios [`v6.1.0`](https://github.com/CardanoSolutions/ogmios/releases/tag/v6.1.0) and cardano-node `8.8.0-pre`. Like before, it is happening with any script version (PlutusV1, V2 and V3). Removing the cardano-node volume did not fixed the issue this time. The error (not being reported) is the same:

> "DecoderErrorDeserialiseFailure \"witness\" (DeserialiseFailure 38 \"expected bytes\")"

I don't have a reproduction repo now as I'm already moving to `v6.2.0` with cardano-node `8.9.1`, but if this version also fails I'll try to provide a repo :+1:.

KtorZ:
Hey, @klarkc I've been trying to reproduce by forcing the an exception at upgrade's call site but the server behaves "as expected":

```
{
  code: -32603,
  message: '"foo"\n' +
    'CallStack (from HasCallStack):\n' +
    '  error, called at src/Relude/Debug.hs:296:11 in rld-1.2.1.0-1d38cbc5:Relude.Debug\n' +
    '  error, called at src/Ogmios/App/Protocol/TxSubmission.hs:296:39 in ogmios-0-inplace:Ogmios.App.Protocol.TxSubmission\n' +
    '  unsafeFromRight, called at src/Ogmios/App/Protocol/TxSubmission.hs:293:28 in ogmios-0-inplace:Ogmios.App.Protocol.TxSubmission'
}
```

Maybe it's the cardano-transaction-lib that isn't handling server errors properly?

KtorZ:
That's a good point. I didn;'t check the server logs (🤦) because I got a response from the client side already.

PR #389, opened 2024-06-03
Update mempool example to v6.x

caike:
Update local-tx-monitor example

### DIFF
--- a/docs/content/mini-protocols/local-tx-monitor.md
+++ b/docs/content/mini-protocols/local-tx-monitor.md
@@ -104,24 +104,27 @@ const client = new WebSocket("ws://localhost:1337");
 // Helper function
 function rpc(method, params) {
     client.send(JSON.stringify({
-        jsonrpc: "2.0",
+        jsonrpc: '2.0',
         method,
         params
     }));
 }
 
 client.on('message', e => {
-  const next = JSON.parse(e).result;
-  if (next.transaction === null) {
-    client.close();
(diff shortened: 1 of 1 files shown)


docs/content/mini-protocols/local-tx-monitor.md
```diff
@@ -104,24 +104,27 @@ const client = new WebSocket("ws://localhost:1337");
...
 }
 
 client.on('message', e => {
-  const next = JSON.parse(e).result;
-  if (next.transaction === null) {
-    client.close();
+  const message = JSON.parse(e);
+  if (message?.result?.transaction === null) {
+    rpc('acquireMempool');
   } else {
```
KtorZ, line 117: I am not sure I understand the rationale behind this change here? Why change the behavior an re-acquire a snapshot?

caike: My understanding is that in the current behavior when the code runs while the mempool is empty, then no transactions will be displayed and `client.close()` will cause the the script to exit. This happened to me when I ran it the first time (preview testnet) and I wasn't sure if I'd done something wrong.

The new code behaves similarly to the [starter kit example](https://github.com/CardanoSolutions/ogmios-ts-client-starter-kit/blob/main/src/local-mempool-monitor.ts#L33-L39) where it loops indefinitely and waits for new transactions to come in.

KtorZ (approved):

PR #406, opened 2024-07-16
Add notes about `stakePools` query being used without params

### LINKED ISSUE #405: Incorrect documentation for `QueryLedgerStateStakePools`
### What Git revision / release tag are you using?

v6.5.0

### Do you use any client SDK? If yes, which one?

None

### Describe what the problem is?

Document states that for `QueryLedgerStateStakePools` endpoint parameter is required, which is not true since `queryLedgerState/stakePools` can be queried without any parameter, in which case, ogmios returns all stake pools in the node. 

### What should be the expected behavior?

Documentation should state that parameter for this endpoint is optional

### If applicable, what are the logs from the server around the occurence of the problem?

_No response_

SeungheonOh:
Closes #405

### DIFF
--- a/docs/content/mini-protocols/local-state-query.md
+++ b/docs/content/mini-protocols/local-state-query.md
@@ -111,7 +111,7 @@ queryLedgerState             | Information
 `proposedProtocolParameters` | The last update proposal w.r.t. protocol parameters, if any.
 `rewardAccountSummaries`     | Current delegation settings and rewards of chosen reward accounts.
 `rewardsProvenance`          | Get details about rewards calculation for the ongoing epoch.
-`stakePools`                 | The list of all currently registered and active stake pools with their current parameters. This query can be made with or without specifying list of target stakePools. When no stake pools are provided in the parameter, query returns list of all stake pools and their parameters in node. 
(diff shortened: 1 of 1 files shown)


docs/content/mini-protocols/local-state-query.md
```diff
@@ -111,7 +111,7 @@ queryLedgerState             | Information
 `proposedProtocolParameters` | The last update proposal w.r.t. protocol parameters, if any.
 `rewardAccountSummaries`     | Current delegation settings and rewards of chosen reward accounts.
 `rewardsProvenance`          | Get details about rewards calculation for the ongoing epoch.
-`stakePools`                 | The list of all currently registered and active stake pools with their current parameters.
+`stakePools`                 | The list of all currently registered and active stake pools with their current parameters. This query can be made with or without specifying list of target stakePools. When no stake pools are provided in the parameter, query returns list of all stake pools and their parameters in node.
```
KtorZ, line 114:
```suggestion
`stakePools`                 | The list of all currently registered and active stake pools with their current parameters. This query can be made with or without specifying list of target `stakePools`. When no stake pools are provided as parameter, the query returns the list of all stake pools and their parameters from the node. 
```

KtorZ (approved):

PR #413, opened 2024-08-16
Fix/docker compose

chrissiwaffler:
- remove the old build instructions (as discussed in discord)
- update the documentation to also download the submodules fully when cloning the repository -> the config submodule is also needed for running the `docker-compose up` command


KtorZ (approved):
Nice 🥰

PR #414, opened 2024-08-22
Fix cost in 7th tier example

mirelon:
It was unclear if there is some rounding involved (padding to 3 decimals in other rows also contributed to this confusion)
For 157.66 the tests in cardano-ledger would fail


KtorZ (approved):
Good point.

PR #419, opened 2024-09-17
Convert raw to string

Sluder:
Resolves the issue `raw.replace is not a function` that seems to randomly occur during indexing. In some cases, `raw` is a buffer, which needs converted to a string before calling the `.replace(...)`


KtorZ (approved):
Nice finding, thanks.

PR #426, opened 2024-12-11
Integrate cardano-node==10.1.3

RitudhwajA:
Bump cardano node to 10.1.3


server/package.yaml
```diff
@@ -1,7 +1,7 @@
 _config: !include ".hpack.config.yaml"
 
 name:                ogmios
-version:             0
+version:             6.10.0
```
KtorZ, line 4: Hey, this one is never bumped on `main`, but only in dedicated branches. Many people build from `main`, and it gets utterly confusing when the version they refer to is some kind of transient version that's not published -- even if the version usually contains a git sha (it doesn't for those building through Nix). 

So anyway, I'll adjust that manually 👍

KtorZ:
Merged manually in 4434bb929

